PR #30, opened 2018-11-26
Validation, completeness and consistency checks before writing netcdf

### LINKED ISSUE #14: Appropriate validation at different stages of dataset creation
There are different levels of validation we can apply at different stages:
* **Schema validation**: making sure each individual dictionary (_dimensions, variables_ and _attributes_) has items that makes sense for its purpose. This is what's currently done with `jsonschema` at template  creation time. It can't be too restrictive because it may not make sense to fully specify all dataset elements in a JSON template (e.g. a generic template could specify the attributes of a TEMP variable, without saying what dimensions it should have).
* **Schema completeness**: Similar to validation, but this level should insist on having _all_ the information required to create a valid netCDF4 object. This could be done using a more strict set of json schemas.
* **Consistency**: This is to ensure that the three dictionaries are consistent with each other; in particular, that the dimensions of variables (and their data arrays) are consistent with what is specified in the _dimensions_ dictionary.

I think validation can be applied at the beginning, but checking for completeness and consistency can be left until just before creating the file. The code is doing something close to this already, but  we need to make sure it's doing the right amount of validation at each stage.

mhidas:
Fixes #14, implementing the apporach described there (https://github.com/aodn/aodn-netcdf-tools/issues/14#issuecomment-438877417)


mhidas:
I've added an [example script](https://github.com/aodn/aodn-netcdf-tools/blob/b888ef8f92724f861df59500b511ae8da61d5583/examples/rottnest.py) to better demonstrate how this works.

mhidas:
@lbesnard @bpasquer @ggalibert @ocehugo @lwgordonimos 
I'm going to merge this so we can build it and it's easier to try it out. Feel free to review and comment anyway.
There's obviously more work to do, but it should start to be useable in a basic way.

mhidas:
You can just `pip install` the wheel from Jenkins: https://jenkins.aodn.org.au/view/ncwriter/job/ncwriter_build/3/artifact/dist/ncwriter-0.0.6-py2-none-any.whl

lwgordonimos:
@mhidas the following mechanism would be more appropriate for checking things like the numpy types:

https://lat.sk/2017/03/custom-json-schema-type-validator-format-python/
https://python-jsonschema.readthedocs.io/en/latest/validate/#jsonschema.FormatChecker

PR #32, opened 2018-11-29
Move schema to a JSON file

### LINKED ISSUE #20: Define schemas in JSON files instead of Python code
https://github.com/aodn/aodn-netcdf-tools/pull/3#discussion_r219022116

e.g. to allow re-use in non-Python applications.

mhidas:
Fixes #20

### DIFF
--- a/ncwriter/schema.py
+++ b/ncwriter/schema.py
@@ -2,6 +2,8 @@
 template, and also the helper functions necessary to validate an object against their respective schema.
 """
 
+import json
+import os
 from copy import deepcopy
 
 import numpy as np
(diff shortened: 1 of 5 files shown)


mhidas:
@lwgordonimos Following your suggestion (https://github.com/aodn/aodn-netcdf-tools/pull/3#discussion_r219022116)

mhidas:
One thing I'm not quite sure about is if we eventually want to use the schema from other code, it should probably be in a separate **schema** directory outside **ncwriter**, but it would still need to be included in the package. Can we still do that in setup.py with the  `package_data` parameter? (e.g. like [here](https://github.com/aodn/python-aodncore/blob/master/setup.py#L54))

ncwriter/schema.py
```diff
@@ -26,77 +28,27 @@
                                       default_types=types
                                       )
 
+TEMPLATE_SCHEMA_JSON = os.path.join(os.path.dirname(__file__), 'template_schema.json')
```
lwgordonimos: Maybe look at the package_data mechanism for including non-Python files?

mhidas: Yep, I'm looking into it.

In terms of accessing the including file using setuptoos functions, [this](https://setuptools.readthedocs.io/en/latest/pkg_resources.html?highlight=data#overview) seems to suggest they rely on having "egg-info" metadata directories. Will that be a problem, given that we're using a wheel to install the package?

mhidas:
See last commit for adding the template JSON file as package data.

Looks like we have the same Travis issue as in aodncore...

PR #33, opened 2018-12-04
.travis.yml: add explicit cython install

mhidas:
Fix Travis cypthon issue, coopying solution from https://github.com/aodn/python-aodncore/pull/141


mhidas:
Looks like this is no longer needed

PR #34, opened 2018-12-04
create custom schema validator without altering the metaschema

mhidas:
Here's my attempt to address https://github.com/aodn/aodn-netcdf-tools/pull/30#issuecomment-443076990
Might need a few more tweaks before merging.

### DIFF
--- a/ncwriter/schema.py
+++ b/ncwriter/schema.py
@@ -5,7 +5,6 @@ template, and also the helper functions necessary to validate an object against 
 import json
 
 import numpy as np
-
 from jsonschema import validators, Draft4Validator, FormatChecker, ValidationError
 from pkg_resources import resource_filename
 
@@ -13,7 +12,6 @@ from pkg_resources import resource_filename
 # Create a new validator class (based on Draft4Validator) to allow templates to use
 # * Python types or numpy dtypes to specify variable data types; and
 # * numpy arrays to specify variable data.
-
 TemplateValidator = validators.create(meta_schema=Draft4Validator.META_SCHEMA,
                                       validators=Draft4Validator.VALIDATORS)
 format_checker = FormatChecker()
@@ -39,10 +37,6 @@ TemplateValidator.check_schema(TEMPLATE_SCHEMA)
 
 template_validator = TemplateValidator(TEMPLATE_SCHEMA, types=TYPES, format_checker=format_checker)
 
-GLOBAL_ATTRIBUTES_SCHEMA = TEMPLATE_SCHEMA.copy()
-GLOBAL_ATTRIBUTES_SCHEMA.pop('properties')  # remove special properties, leaving only global attributes
-global_attribute_validator = TemplateValidator(GLOBAL_ATTRIBUTES_SCHEMA, types=TYPES, format_checker=format_checker)
-
 
 def validate_template(t):
     template_validator.validate(t)
@@ -57,4 +51,8 @@ def validate_variables(v):
 
 
 def validate_global_attributes(a):
-    global_attribute_validator.validate(a)
+    if hasattr(a, 'keys'):
+        special = [k for k in a.keys() if k.startswith('_')]
+        if special:
+            raise ValidationError('Special attributes {} not allowed in global attributes dict'.format(special))
+    template_validator.validate(a)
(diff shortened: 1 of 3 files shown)


ncwriter/schema.py
```diff
@@ -18,28 +17,53 @@
...
+    """Return whether the given value is a valid data type specification for a NetCDF variable"""
+    if isinstance(value, basestring):
+        if value in TYPE_STRINGS:
+            return True
+        else:
+            return False
+    if isinstance(value, np.dtype):
+        return True
+    if isinstance(value, type):
+        if issubclass(value, np.number):
```
lwgordonimos: Already a bool function, so this can just be

```python
return issubclass(value, np.number)
```

Also, this suggests that sometimes it is receiving an *instance* and sometimes it is receiving a *class* reference? What's the rationale behind that?

ncwriter/schema.py
```diff
@@ -18,28 +17,53 @@
...
+                'S1', 'c', 'i1', 'b', 'B', 'u1', 'i2', 'h', 's', 'u2', 'i4', 'i', 'l', 'u4',
+                'i8', 'u8', 'f4', 'f', 'f8', 'd'
+                )
+
+
+@format_checker.checks('datatype')
+def is_datatype(value):
+    """Return whether the given value is a valid data type specification for a NetCDF variable"""
+    if isinstance(value, basestring):
+        if value in TYPE_STRINGS:
```
lwgordonimos: Already a boolean expression, so can just be:
```python
return value in TYPE_STRINGS
```

ncwriter/schema.py
```diff
@@ -18,28 +17,53 @@
...
 
 TEMPLATE_SCHEMA_JSON = resource_filename(__name__, 'template_schema.json')
 with open(TEMPLATE_SCHEMA_JSON) as f:
     TEMPLATE_SCHEMA = json.load(f)
 TemplateValidator.check_schema(TEMPLATE_SCHEMA)
 
+template_validator = TemplateValidator(TEMPLATE_SCHEMA, types=TYPES, format_checker=format_checker)
 
 GLOBAL_ATTRIBUTES_SCHEMA = TEMPLATE_SCHEMA.copy()
 GLOBAL_ATTRIBUTES_SCHEMA.pop('properties')  # remove special properties, leaving only global attributes
```
lwgordonimos: I feel like there has to be a native way to handle this without a copy/pop but I don't know what that is! This is the only instance of fiddling with the JSON from Python, so it may become clear in future how to manage this...

ncwriter/schema.py
mhidas: :+1: Ha, of course! That's the sort of thing I've pointed out before when reviewing other people's code, but it's different when you're writing it yourself...

Actually I'm thinking of replacing the string checking here with an `enum` type in the schema

ncwriter/schema.py
mhidas: This is validating values that can be passed to the `datatype` arg of netCDF [`createVariable`](http://unidata.github.io/netcdf4-python/#netCDF4.Dataset.createVariable) method. The docs say:
> The datatype can be a numpy datatype object, or a string that describes a numpy dtype object (like the dtype.str attribute of a numpy array). Supported specifiers include: 'S1' or 'c' (NC_CHAR), 'i1' or 'b' or 'B' (NC_BYTE), 'u1' (NC_UBYTE), 'i2' or 'h' or 's' (NC_SHORT), 'u2' (NC_USHORT), 'i4' or 'i' or 'l' (NC_INT), 'u4' (NC_UINT), 'i8' (NC_INT64), 'u8' (NC_UINT64), 'f4' or 'f' (NC_FLOAT), 'f8' or 'd' (NC_DOUBLE)

So that means _instances_ of `numpy.dtype`, or the strings listed. From experimenting, I know it can also accept more informative strings like 'float32', 'int64', etc..., as well as _classes_ such as `numpy.float32`, which are subclasses of `numpy.number`.

So to maintain this flexibility, we need to allow for all these possibilities.

mhidas: Alternatively, I could just not validate to this level, and allow the netCDF4 library to complain about anything it can't make sense of. That would simplify the verification code here, but in order to give more helpful error messages, we'd have to catch these exceptions and add more info, so there would be more overhead in the `DatasetTemplate` methods that write the file.

ncwriter/schema.py
mhidas: Yeah, I know it's a bit hacky. I do have a couple of ideas for avoiding this. I'll try them out.

Issue #19, opened 2018-10-15
Do we really need 4 fill value aliases?


mhidas:
Currently the code allows the fill value for a variable to be specified in an attribute with any of these names 
```python
self.fill_aliases = {'fill_value', 'missing_value', 'FillValue', '_FillValue'}
```
It doesn't matter which you use, this is treated as a special attribute and passed as the `fill_value` option to [`Dataset.createVariable()`](http://unidata.github.io/netcdf4-python/#netCDF4.Dataset.createVariable). The  actual attribute name written to the netCDF file is always  "_FillValue", which is the standard one specified by [CF](http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#missing-data) and the NetCDF User Guide ([NUG](https://www.unidata.ucar.edu/software/netcdf/docs/attribute_conventions.html)) (Both of these also mention "missing_value", but it has a slightly different definition, is allowed to be a vector, and at least according to CF [Appendix A](http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#attribute-appendix) it is actually deprecated by the NUG.)

So to avoid confusion and simplify the code, I think it would make sense to only treat "_FillValue" as a special attribute, i.e. it should be the way to specify what the fill value should be.

mhidas:
"missing_value" may have a similar meaning, but is **not equivalent** to "_FillValue", according to the [NUG](https://www.unidata.ucar.edu/software/netcdf/docs/attribute_conventions.html)
> _FillValue
> The _FillValue attribute specifies the fill value used to pre-fill disk space allocated to the variable. Such pre-fill occurs unless nofill mode is set using nc_set_fill(). The fill value is returned when reading values that were never written. If _FillValue is defined then it should be scalar and of the same type as the variable.

> missing_value
This attribute **is not treated in any special way by the library or conforming generic applications**, but is often useful documentation and may be used by specific applications. The missing_value attribute **can be a scalar or vector** containing values indicating missing data.

Based on that, I don't think we should treat an attribute named "missing_value" in any special way.

I think we should only accept "_fill_value" or "_FillValue"

mhidas:
Fixed in #36

PR #36, opened 2018-12-18
Handle fill value aliases

### LINKED ISSUE #27: raise a warning when touching more than one FillValues aliases
There should be an explicit warning when user provide two or more _FillValue aliases, given a silent overwrite of the _FillValue is being done and they can be quite different and change the final outcomes.

mhidas:
Fixes #27

### DIFF
--- a/ncwriter/template.py
+++ b/ncwriter/template.py
@@ -188,8 +188,7 @@ class DatasetTemplate(NetCDFGroupDict):
             if "_datatype" not in var:
                 datatype = getattr(var["_data"], 'dtype', None)
                 if datatype is not None:
-                    print("WARNING: Guessed data type {datatype} "
-                          "for variable '{name}'".format(datatype=datatype, name=name))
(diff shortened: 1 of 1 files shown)


mhidas:
This fixes #19 as well, as per my [suggestion](https://github.com/aodn/aodn-netcdf-tools/issues/19#issuecomment-444761366).

ncwriter/template.py
```diff
@@ -247,30 +248,38 @@ def ensure_consistency(self):
...
-            raise ValueError('You can only provide one missing value alias!')
+            fill_dict = {f: special_dict[f] for f in fill_aliases}
+            unique_values = set(fill_dict.values())
+            msg = "Multiple fill value options specified for variable '{vname}': {fill_dict}.".format(
+                vname=vname, fill_dict=fill_dict
+            )
+            if len(unique_values) == 1:
+                warn(msg)
+            else:
+                raise ValueError(msg)
```
ggalibert: You  might want to make it clear that the problem is that there are multiple distinct values for fill value aliases.

Issue #16, opened 2018-10-11
IMOS template


mhidas:
Create a subclass of `DatasetTemplate` called `ImosTemplate`, which
* Automatically includes the fixed global attributes required by the IMOS convention;
* Can also include IMOS standard variable attributes (standard names, units, fill values, e.g.  as specified in imosParameters.txt in imos-toolbox)?
* Has a helper method to auto-generate an IMOS-compliant file name for a template (based on time coverage, platform_code, etc...,  see e.g.  [here](https://github.com/aodn/data-services/blob/024a26e7ceec2334f34a5a19dd237b85e065afea/lib/python/IMOSnetCDF.py#L237))
* Add a `date_created` attribute when creating the netCDF file.

mhidas:
IMOS template class created in #39, including global attributes and date_created. I've opened separate issues (#40 & #41) for the remaining functionality to be added later.

PR #39, opened 2019-04-03
IMOS template

### LINKED ISSUE #17: Automatically calculate temporal/spatial extent attributes
Once a template has been fully completed with data arrays, the code could use the ranges in the relevant time and spatial variables to calculate the values of the global attributes `time_coverage_min/max`, `geospattial_lat_min/max` etc..., as done e.g. [here](https://github.com/aodn/data-services/blob/024a26e7ceec2334f34a5a19dd237b85e065afea/lib/python/IMOSnetCDF.py#L168)

mhidas:
Create an ImosTemplate subclass to simplify creating files that meet the IMOS-1.4 conventions.

Fixes #17 and part of #16.


mhidas:
See the difference between these two scripts producing the same CF- & IMOS-compliant file:
* [rottnest.py using just DatasetTemplate](https://github.com/aodn/aodn-netcdf-tools/blob/9b0e7bf8787579a6a5e4a9260faf34fa451994f2/examples/rottnest.py);
* [rottnest.py using ImosTemplate](https://github.com/aodn/aodn-netcdf-tools/blob/imos_template/examples/rottnest.py).

mhidas:
I'm going to merge this because I'd like to make  use of  it in the pipelines. It only introduces new functionality, without affecting existing behaviour (there is one new method added to `DatasetTemplate`).

Feel free to post-review.

PR #42, opened 2019-04-24
Jenkinsfile For aodn-netcdf-tools

pmbohm:
for aodn/backlog#1118

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,23 @@
+pipeline {
+    agent { dockerfile true }
+
+    stages {
+        stage('test') {
+            steps {
+                sh 'python setup.py test'
+            }
+        }
+        stage('package') {
+            steps {
+                sh 'python setup.py bdist_wheel'
+            }
+        }
+    }
+    post {
+        success {
+            dir('dist/') {
+                archiveArtifacts artifacts: '*.whl', fingerprint: true, onlyIfSuccessful: true
+            }
+        }
+    }
+}
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,22 +1,51 @@
+#!groovy
+
 pipeline {
-    agent { dockerfile true }
+    agent none
 
     stages {
-        stage('test') {
+        stage('clean') {
+            agent { label 'master' }
             steps {
-                sh 'python setup.py test'
+                sh 'git clean -fdx'
             }
         }
-        stage('package') {
-            steps {
-                sh 'python setup.py bdist_wheel'
+
+        stage('container') {
+            agent {
+                dockerfile {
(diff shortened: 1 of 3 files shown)


akashisama:
@lwgordonimos bumpversion seems to bet setting it to the same version (0.2.1.dev0) for every build? See my 3 builds at http://jenkins-alexs-fresh-jenkins.dev.aodn.org.au/job/aodn-netcdf-tools_build/job/Jenkinsfile/

akashisama:
Ah ok just noticed the "when { branch 'master' }" line, that's presumably where I was going wrong

akashisama:
@lwgordonimos I created a branch that set itself as the master and had the same problem. I think there needs to be a git push after "bumpversion --tag --commit release" so the version change is kept track of?

http://jenkins-alexs-fresh-jenkins.dev.aodn.org.au/job/aodn-netcdf-tools_build/job/alexs-jenkinsfile/

lwgordonimos:
Yeah, I didn't want to interact with Github during development *too* much.

In prod it will need something like:
`git push origin --tags`

Mentions it here: https://github.com/peritus/bumpversion

```
tag = (True | False)

    default: False (Don't create a tag)

    Whether to create a tag, that is the new version, prefixed with the character "v".
If you are using git, don't forget to git-push with the --tags flag.
```

.bumpversion.cfg
```diff
@@ -0,0 +1,18 @@
+[bumpversion]
+current_version = 0.2.0
+commit = False
+tag = False
+parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(\.(?P<release>[a-z]+)(?P<build>\d+))?
```
akashisama: including the build number seems like a good idea, but the 'build' part here seems to always be '0'

Jenkinsfile
```diff
@@ -0,0 +1,55 @@
...
+            stages {
+                stage('version') {
+                    steps {
+                        sh 'bumpversion patch'
+                    }
+                }
+                stage('release') {
+                    when { branch 'master' }
+                    steps {
+                        sh 'bumpversion --tag --commit release'
```
akashisama: git will need to be installed in the docker to make this work

lwgordonimos: Ah yeah, good point. We might need to refactor it to run outside the container, since it just operates against the git checkout, and isn't actually part of the build toolchain anyway.

akashisama:
I'm a bit worried about merging this since the versioning changes will break the current jenkins build/release process...

lwgordonimos:
Yes, I was going to mention that this morning. Start with the non-Python projects first, and we might need to pull the version related stuff out and deal with it when we look at releases.

PR #45, opened 2019-06-07
Set date_created attribute to UTC time, not local time

mhidas:


mhidas:
I just noticed this was being incorrectly set to the local time, then recorded in the global attribute as if it was UTC.

PR #49, opened 2019-08-13
Refactor package to include products code

mhidas:
This is in preparation for adding all the product-generation code into this package, so it can be separate from aodndata (which is pipeline-specific).

This repo will be renamed to **python-aodntools**.


mhidas:
@lwgordonimos I think we'll also need to get rid of versioneer, and switch to the new way of managing package versions.

mhidas:
@lwgordonimos I have rebased this onto the "bumpversion" version and tried to fix up the new config files to match the new package structure. Hope this works.

aodntools/ncwriter/imos_template.py
```diff
@@ -3,8 +3,7 @@
 
 from netCDF4 import num2date
 
-from ncwriter import DatasetTemplate
-
+from aodntools.ncwriter import DatasetTemplate
```
lwgordonimos: The internal package imports are all relative except this one. Maybe should be:

```python
from .template import DatasetTemplate
```

?

mhidas: :+1: Done

mhidas:
Thanks @lwgordonimos . I've fixed that relative import, and added the config.cfg file.

PR #50, opened 2019-08-16
Replace versioneer with bumpversion pattern

anguss00:


lwgordonimos:
@mhidas This overlaps with your comment in https://github.com/aodn/aodn-netcdf-tools/pull/49

Are you happy to merge this before your refactoring one goes through? If so, we'll just have to update the Jenkins job definition to suit (which is trivial).

mhidas:
:+1: @lwgordonimos Yep, go for it!

PR #63, opened 2019-10-07
Hourly timeseries

diodon:
modify the output file name and the global to handle products with qcflags 0,1,2 and qcflags 1,2

### DIFF
--- a/aodntools/timeseries_products/hourly_timeseries.py
+++ b/aodntools/timeseries_products/hourly_timeseries.py
@@ -535,11 +535,15 @@ def hourly_aggregator(files_to_aggregate, site_code, qcflags, file_path ='./'):
 
     ## add global attributes
     add_attribute = {'rejected_files': "\n".join(list(bad_files)),
-                     'values_retained':  ", ".join([qcflags_names[flag] for flag in qcflags])}
+                     'included_values_flagged_as':  ", ".join([qcflags_names[flag] for flag in qcflags])}
     nc_aggregated.attrs = set_globalattr(nc_aggregated, TEMPLATE_JSON, site_code, add_attribute, parameter_names)
-    nc_aggregated.attrs['abstract'] += 'Only values flagged as ' + ", ".join([qcflags_names[flag] for flag in qcflags]) + ' are retained in the aggregation'
+    nc_aggregated.attrs['abstract'] = nc_aggregated.attrs['abstract'].format(
+        site_code=site_code,
+        flags=", ".join(qcflags_names[flag] for flag in qcflags)
+    )
     if 0 in qcflags:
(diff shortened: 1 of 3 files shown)


aodntools/timeseries_products/hourly_timeseries.py
```diff
@@ -426,6 +426,8 @@ def hourly_aggregator(files_to_aggregate, site_code, qcflags, file_path ='./'):
     parameter_names_accepted = ['DEPTH', 'CPHL', 'CHLF', 'CHLU', 'DOX', 'DOX1', 'DOX1_2', 'DOX1_3', 'DOX2',
                                 'DOX2_1', 'DOXS', 'DOXY', 'PRES', 'PRES_REL', 'PSAL', 'TEMP', 'TURB', 'PAR']
     function_stats = ['min', 'max', 'std', 'count']
+    qcflags_names = {0: 'Non_qc_performed', 1: 'Good_data', 2: 'Probably_good_data',
```
mhidas: 0 is "No_QC_performed"

aodntools/timeseries_products/hourly_timeseries.py
```diff
@@ -595,11 +595,9 @@ def hourly_aggregator(files_to_aggregate, site_code, qcflags, file_path ='./'):
     ncout_filename = generate_netcdf_output_filename(nc=nc_aggregated, facility_code=facility_code, data_code=data_code,
                                                      site_code=site_code,
                                                      product_type=product_type, file_version=file_version)
-    ncout_path = os.path.join(file_path, ncout_filename)
-
-    write_netCDF_aggfile(nc_aggregated, ncout_path, encoding, file_path)
```
mhidas: To me it makes more sense the way you had it before. Just create `ncout_path` _once_ by doing the join, then that's all you need to supply to `write_netCDF_aggfile`. That function doesn't separately need to know what the `file_path` is.

aodntools/timeseries_products/hourly_timeseries.py
```diff
@@ -337,7 +337,7 @@ def generate_netcdf_output_filename(nc, facility_code, data_code, site_code, pro
 
 
 
-def write_netCDF_aggfile(nc_aggregated, ncout_filename, encoding, file_path):
+def write_netCDF_aggfile(nc_aggregated, ncout_filename, encoding):
     """
     write netcdf file
     :param file_path: path where to write the file
```
mhidas: Might as well update the doc string to be consistent with the function arguments...

aodntools/timeseries_products/hourly_timeseries.py
mhidas: :+1:

aodntools/timeseries_products/hourly_timeseries.py
diodon: done

PR #79, opened 2019-11-06
Hourly LTSP: add contributors global attr

diodon:
add contributors name, emall and role to the global attributes of the resulting file

### DIFF
--- a/aodntools/timeseries_products/hourly_timeseries.py
+++ b/aodntools/timeseries_products/hourly_timeseries.py
@@ -11,7 +11,7 @@ import pandas as pd
 import xarray as xr
 from dateutil.parser import parse
 from pkg_resources import resource_filename
-from aggregated_timeseries import get_contributors
+from aodntools.timeseries_products.aggregated_timeseries import get_contributors
 
 
 TEMPLATE_JSON = resource_filename(__name__, 'hourly_timeseries_template.json')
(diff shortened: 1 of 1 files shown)


aodntools/timeseries_products/hourly_timeseries.py
```diff
@@ -217,6 +217,35 @@ def get_nominal_depth(nc):
     return nominal_depth
 
 
+def get_contributors(files_to_agg):
```
mhidas: This is exactly the same code you already have in `aggregated_timeseries.py`, so it is not necessary to repeat it here. All you have to do is import it (at the top of the file):
```python
from aodntools.timeseries_products.aggregated_timeseries import get_contributors
```

diodon: done. Perhaps you need to adjust the path of the aggregated_timeseries package.

mhidas: Actually, what I suggested above was correct, as an absolute import. Alternatively, you need to explicitly make it a relative import with a leading '.' (see [PEP 328](https://www.python.org/dev/peps/pep-0328)):
```python 
from .aggregated_timeseries import get_contributors
```
Currently, the unittest fails because it can't find the module: https://travis-ci.org/aodn/python-aodntools/jobs/611674191#L299

PR #83, opened 2019-11-19
Gridded timeseries

diodon:
interpolate TEMP or PSAL into fixed depth bins from the hourly timeseries product


diodon:
I will close it as the code changed significantly and no one has review this version so far...

PR #86, opened 2019-12-09
Remove Python2 compatibility and unpin requirement versions

mhidas:


aodntools/timeseries_products/aggregated_timeseries.py
```diff
@@ -1,4 +1,5 @@
-from __future__ import print_function
+#! /usr/bin/env python3
```
lwgordonimos: There isn't usually a space before the interpreter path

mhidas: :+1: Fixed

mhidas:
This should be good to merge too.

PR #87, opened 2019-12-10
bump version to 1.0

mhidas:
This should have been part of #86


lwgordonimos:
This has come through with a dev version number still.

We'll document this better, but try the following to bump it to a "release" version (i.e. without the dev0):

```bash
bump2version major
bump2version --commit --allow-dirty release
```

mhidas:
:+1: Done.

PR #89, opened 2019-12-19
aggregated time series - pure numpy version

diodon:
This code produces exactly the same aggregated-timeseries product using aprox. half of the memory resources, and it is a little bit faster. Also, the data variables are compressed, generating a compact output file.


mhidas:
@diodon I assume #97 makes this PR redundant, so we can close it?

diodon:
Yes.please close it.

PR #90, opened 2020-01-12
Gridded timeseries

diodon:
grid TEMP or PSAL values from hourly timeseries product into fixed target depths.


aodntools/timeseries_products/gridded_timeseries.py
```diff
@@ -0,0 +1,260 @@
...
+import xarray as xr
+import pandas as pd
+
+import aggregated_timeseries as TStools
+
+
+TEMPLATE_JSON = 'gridded_timeseries_template.json'
+
+
+def make_depth_bins(nc, increment=10):
```
ocehugo: I would prefer to have DEPTH as argument instead of ```nc```.

aodntools/timeseries_products/gridded_timeseries.py
```diff
@@ -0,0 +1,260 @@
...
+def adjust_depth_bins(depth_bins, depth_min, depth_max):
+    """
+    Adjust the provided depth bins to match the min and max registered depth
+    :param depth_bins: list of target depths
+    :param depth_min: min valid depth
+    :param depth_max: max valid depth
+    :return: list of adjusted target depths
+    """
+
+    target_depth_min = bisect.bisect_left(depth_bins, depth_min)
```
ocehugo: You use bisect only two times and for very simple stuff - finding the closest index of a depth value.
IMO, I don't think that using an external dependency for that is required.

Although the numpy way is a bit verbose, a bisect function is a one liner:

```
def bisect_left(alist,value):
return np.where(np.array(x)<=value)[0][-1]
```

bonus are: faster and one less dependency.

aodntools/timeseries_products/gridded_timeseries.py
```diff
@@ -0,0 +1,260 @@
...
+    depth_mask = [False] * len(depth_bins)
+    for depth_index in range(len(depths) - 1):
+        if depths[depth_index + 1] - depths[depth_index] < max_separation:
+            depth_min = bisect.bisect_left(depth_bins, depths[depth_index])
+            depth_max = bisect.bisect_right(depth_bins, depths[depth_index + 1])
+            depth_interval_len = depth_max - depth_min
+            depth_mask[depth_min:depth_max] = [True] * depth_interval_len
+    return depth_mask
+
+def sort_attributes(attributes):
```
ocehugo: I don't see any logic that requires a sorted dictionary - So why do it? If it's for the final file to have global attributes in order, you can do at write time.

aodntools/timeseries_products/gridded_timeseries.py
```diff
@@ -0,0 +1,260 @@
...
+    """
+
+    :param attributes: dictionary of global attributes
+    :return: sorted dictionary
+    """
+    return dict(sorted(attributes.items()))
+
+def sort_depths(depths, values):
```
ocehugo: You are mutating the arguments, this is not recommended.

Also, I don't think your argument types (lists) are the best. In this case arrays are much better since you are doing index operations.

If you use numpy and are happy in using arrays, a much cleaner version is possible:

```
sorted_indexes = np.argsort(depths)
sorted_depths = depths[sorted_indexes]
sorted_values = values[sorted_indexes]
```

aodntools/timeseries_products/gridded_timeseries.py
```diff
@@ -0,0 +1,260 @@
...
+    :param VoI: variable of interest (TEMP or PSAL)
+    :param depth_bins: list of depth where to interpolate. if null list is provided it will be calculated from the data
+    :param max_separation: max separation allowed for instruments
+    :param depth_bins_increment: in case no depth bins provided this is the increment for the calculated bins
+    :param base_path: path where the result file will be written
+    :return: interpolated dataset
+    """
+
+    with xr.open_dataset(file_name) as nc:
+        site_code = nc.site_code
```
ocehugo: ask for it, otherwise, you will have keyerror.

aodntools/timeseries_products/gridded_timeseries.py
```diff
@@ -0,0 +1,260 @@
...
+        ## get global attributes
+        global_attributes = nc.attrs
+
+        ## get lat/lon
```
ocehugo: same as above

aodntools/timeseries_products/gridded_timeseries.py
```diff
@@ -0,0 +1,262 @@
...
+    :param VoI: variable of interest (TEMP or PSAL)
+    :param depth_bins: list of depth where to interpolate. if null list is provided it will be calculated from the data
+    :param max_separation: max separation allowed for instruments
+    :param depth_bins_increment: in case no depth bins provided this is the increment for the calculated bins
+    :param base_path: path where the result file will be written
+    :return: interpolated dataset
+    """
+
+    with xr.open_dataset(file_name) as nc:
+        site_code = nc.site_code
```
ocehugo: If you must, all those dictionary access could be checked before hand and hence, bail out before computing stuff - for example:

```
required_keys = ['site_code','LONGITUDE','LATITUDE','etc']
kerror = [f"key={key} not found\n") for key in nc.keys() if key not in required_keys]
if kerror:
raise KeyError(''.join(kerror))
end
```

You could even wrap that in a function to be called before everything else so you know beforehand which files are incomplete.

aodntools/timeseries_products/gridded_timeseries.py
```diff
@@ -0,0 +1,262 @@
...
+            depth_bins = make_depth_bins(nc=nc, increment=depth_bins_increment)
+        else:
+            depth_bins = [float(depth) for depth in depth_bins]     # in case depth_bins provided through inline arguments
+            depth_bins = adjust_depth_bins(depth_bins, nc.DEPTH.min(), nc.DEPTH.max())
+
+        ## create empty containers
+        time_0 = pd.to_datetime('1950-01-01T00:00:00')
+        time_min = nc.TIME.values.min()
+        depth_bin_len = len(depth_bins)
+        VoI_binned = xr.DataArray(np.full((depth_bin_len, 1), np.nan), coords=[[time_0], depth_bins],
```
ocehugo: maybe is of interest to just check the size before allocating - if depth_bin_len is too big you may explode memory usage.

aodntools/timeseries_products/gridded_timeseries.py
```diff
@@ -0,0 +1,262 @@
...
+                                  dims=['TIME', 'DEPTH'])
+        VoI_ndepths = xr.DataArray(np.full(1, 0), coords=[[time_0]], dims=['TIME'])
+
+        ## group nc by individual timestamps
+        VoI_grouped = nc.groupby('TIME')
+
+
+        for timestamp, (name, group) in enumerate(VoI_grouped):
```
ocehugo: this for loop is too big. the case for n_depths>=2 should be a function.

aodntools/timeseries_products/gridded_timeseries.py
```diff
@@ -0,0 +1,262 @@
...
+                VoI_gridded = np.full((depth_bin_len, 1), np.nan)
+
+            VoI_temp = xr.DataArray(VoI_gridded.reshape(depth_bin_len, 1), coords=[time, depth_bins],
+                                    dims=['TIME', 'DEPTH'])
+            VoI_ndepths_temp = xr.DataArray([n_depths], coords=[time], dims=['TIME'])
+
+            VoI_binned = xr.concat([VoI_binned, VoI_temp], dim='TIME')
+            VoI_ndepths = xr.concat([VoI_ndepths, VoI_ndepths_temp], dim='TIME')
+
+    VoI_interpolated = xr.Dataset({VoI: VoI_binned.astype(float),
```
ocehugo: do the conversion to float earlier so all operations are faster.

aodntools/timeseries_products/gridded_timeseries.py
```diff
@@ -0,0 +1,262 @@
...
+                                   VoI + '_count': VoI_ndepths.astype(int)})
+
+    ## drop the very first record as it is dummy
+    VoI_interpolated = VoI_interpolated.where(VoI_interpolated.TIME >= time_min, drop=True)
```
ocehugo: this overwrites could be a function

aodntools/timeseries_products/gridded_timeseries.py
```diff
@@ -0,0 +1,262 @@
...
+            if n_depths >= 2:
+                VoI_values = list(group[VoI].values)
+                depth = list(group.DEPTH.values)
+                ## sort depths
+                depth, VoI_values = sort_depths(depth, VoI_values)
+
+                ## check for max separation
+                depth_mask = get_depth_mask(depth_bins=depth_bins, depths=depth, max_separation=max_separation)
+                ## do the interpolation
+                VoI_gridded = np.interp(depth_bins, depth, VoI_values, left=np.nan, right=np.nan)
```
ocehugo: you can probably store the weights and re-use them if this loop is more than 2 lengths.

Also, be aware that np.interp do not check bounds.

aodntools/timeseries_products/gridded_timeseries.py
```diff
@@ -0,0 +1,262 @@
...
+        for timestamp, (name, group) in enumerate(VoI_grouped):
+            time = [name]
+            n_depths = int(len(group[VoI]))
+
+            if n_depths >= 2:
+                VoI_values = list(group[VoI].values)
+                depth = list(group.DEPTH.values)
+                ## sort depths
+                depth, VoI_values = sort_depths(depth, VoI_values)
+
```
ocehugo: Again - your choice of using lists instead of arrays is creating more work - you convert an array to list (```depth```) then you mutate a list in sort_depths with e a more verbose method. You are saved because list(group.DEPTH.values) copies the values but this could be simpler with just 

```sort_depths(group.DEPTH.values,VoI_values)```

File: aodntools/timeseries_products/gridded_timeseries.py [hunk 14]
```diff
@@ -0,0 +1,262 @@
...
+                ## check for max separation
+                depth_mask = get_depth_mask(depth_bins=depth_bins, depths=depth, max_separation=max_separation)
+                ## do the interpolation
+                VoI_gridded = np.interp(depth_bins, depth, VoI_values, left=np.nan, right=np.nan)
+                ## set masked depth bins to zero
+                VoI_gridded = VoI_gridded * depth_mask
+                VoI_gridded[VoI_gridded == 0] = np.nan
+
```
ocehugo: I would think about the interp operation here - maybe you can split the interpolation just to the valid ranges to reduce memory footprint plus avoiding the above masking bits.

aodntools/timeseries_products/gridded_timeseries.py
ocehugo: Also, if you are masking based on time_min, it's memory wise to apply that before interpolation.

aodntools/timeseries_products/gridded_timeseries.py
```diff
@@ -0,0 +1,262 @@
...
+    ## drop the very first record as it is dummy
+    VoI_interpolated = VoI_interpolated.where(VoI_interpolated.TIME >= time_min, drop=True)
+
+    ## Add lat/lon as scalar variables
+    VoI_interpolated = VoI_interpolated.assign(LONGITUDE = longitude_mean,
+                                               LATITUDE = latitude_mean)
+
+    ## transpose dimensions to make CF compliant
+    VoI_interpolated = VoI_interpolated.transpose('TIME', 'DEPTH')
```
ocehugo: avoid the transpose by putting the data in the expected shape before hand.

aodntools/timeseries_products/gridded_timeseries.py
```diff
@@ -0,0 +1,262 @@
...
+
+    ## get the variables attribute dictionary
+    with open(TEMPLATE_JSON) as json_file:
+        attr_dictionary = json.load(json_file)
+
+    variable_attribute_dictionary = attr_dictionary['_variables']
+    global_attribute_dictionary = attr_dictionary['_global']
+
+    ## set variable attributes
```
ocehugo: this can be a function "set_variable-attributes"

aodntools/timeseries_products/gridded_timeseries.py
```diff
@@ -0,0 +1,262 @@
...
+    varlist = list(VoI_interpolated.variables)
+    add_variable_attribute = {}
+    variable_attributes = TStools.set_variableattr(varlist, variable_attribute_dictionary, add_variable_attribute)
+    time_units = variable_attributes['TIME'].pop('units')
+    time_calendar = variable_attributes['TIME'].pop('calendar')
+    for variable in varlist:
+        VoI_interpolated[variable].attrs = variable_attributes[variable]
+
+    ## set global attributes
```
ocehugo: same as above

aodntools/timeseries_products/gridded_timeseries.py
```diff
@@ -0,0 +1,262 @@
...
+    timeformat = '%Y-%m-%dT%H:%M:%SZ'
+    VoI_interpolated.attrs.update(global_attributes)
```
ocehugo: I'm a purist - if you are overwriting, just overwrite - the update method is just noise here. 

Also, it's very hard to know what is being updated  because the argument is enormous and the variable names you created are big (e.g. global_attribute_dictionary).
Since their scope is short - you only use them here - a short name is handy.

finally, it's much easier to update a key at a time, and/or to create a dictionary before hand:

```
dt_start_str = pd.to_datetime(VoI_interpolated.TIME.values.min()).strftime(timeformat);
...
updated_gattrs = {'file_version': gatt_dict['file_version'],
                  'source_file': file_name,
                  'time_coverage_start': dt_start_str,
...
```

aodntools/timeseries_products/gridded_timeseries.py
```diff
@@ -0,0 +1,262 @@
...
+    encoding = {'TIME': {'_FillValue': None,
+                         'units': time_units,
+                         'calendar': time_calendar},
+                VoI+'_count': {'_FillValue': None,
+                               'dtype': np.dtype('int')},
+                'DEPTH': {'dtype': np.dtype('float64')},
+                'LONGITUDE': {'_FillValue': False},
+                'LATITUDE': {'_FillValue': False}}
+
+    TStools.write_netCDF_aggfile(VoI_interpolated, ncout_path, encoding)
```
ocehugo: maybe compression is missing here?

aodntools/timeseries_products/gridded_timeseries.py
```diff
@@ -0,0 +1,262 @@
...
+    parser.add_argument('-file', dest='filename', help='name of the Hourly Time Series Product file that contains the data', default=None, required=False)
+    parser.add_argument('-depth_bins', dest='depth_bins', help='list of depth where the VoI will be interpolated', default=None, nargs='+', required=False)
+    parser.add_argument('-max_separation', dest='max_separation', help='maximum difference between instruments to allow interpolation', default=50, required=False)
+    parser.add_argument('-depth_bins_increment', dest='depth_bins_increment', help='increment in meters for the automatic generated depth bins', default=10, required=False)
+    parser.add_argument('-path', dest='output_path', help='path where the result file will be written. Default ./', default='./', required=False)
+    parser.add_argument('-config', dest='config_file', help='JSON configuration file', default=None, required=False)
+    args = parser.parse_args()
+
+    if args.config_file:
+        with open(config_file) as ff:
```
ocehugo: the if logic here is very similar, I would recommend reviewing the key/value names (i.e. using "output_path" or "path" for both cases.

You are also using dict like access and method access, which is just a wart of using different classes/traits - maybe if you read one as dictioanry the block below will be simpler and more elegant.

ocehugo:
@diodon , comments up for grabs!

aodntools/timeseries_products/gridded_timeseries.py
diodon: I've experimented with bisect and custom made bisect functions as you suggested, but the bisect package resulted way more faster that the custom made bisect functions:

with the `bisect` package
```
In [21]: %timeit get_depth_mask(depth_bins, depths, max_separation)
2.16 µs ± 19.7 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```
with the custom-made functions
```
In [20]: %timeit get_depth_mask2(depth_bins, depths, max_separation)
15.9 µs ± 178 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

```

ocehugo: I read the units wrongly then. I just notice that its from the standard library - so disregard the dependency and the comment above

aodntools/timeseries_products/gridded_timeseries.py
diodon: yes, is to have the global attrs sorted in the output file. the sort in now done just before writing without the function

aodntools/timeseries_products/gridded_timeseries.py
diodon: fixed mutation of the arguments. 
Using numpy array is slower...

aodntools/timeseries_products/gridded_timeseries.py
diodon: not needed as we're using the hourly aggregated times series product as input, which is completely under our control.

aodntools/timeseries_products/gridded_timeseries.py
ocehugo: The type makes a difference:

```
def sort_values_by_depth(depth,values)
       x=np.argsort(depth)
       return depth[x],values[x]
```

The above take 20us here.
The list code takes 130-150us and it's also longer and more verbose.

If I need to convert types, then the above code will take 170-190us, not much more for a shorter and less verbose code:

```
def sort_values_by_depth(depth,values)
       x=np.argsort(depth)
       return np.array(depth)[x], np.array(values)[x]
```

aodntools/timeseries_products/gridded_timeseries.py
diodon: True, but we're not expecting more than ~100 target depths. Even 1000 target depth should be manageable.

aodntools/timeseries_products/gridded_timeseries.py
diodon: I agree that the numpy approach is cleaner but I cannot replicate your results. My tests showed that the list-approach sorting is almost 3 times faster that the np.array approach in small lists. Longer the list, smaller the difference, but we're not expecting to process lists longer that ~100 elements:

list approach:
```
depth = list(range(0,100, 10))+[15, 98, 43]
values = list(range(len(depth)))

In [10]: %timeit sort_depths(depth, values)
2.45 µs ± 48.1 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```
np.array approach:
```
In [13]: %timeit sort_values_by_depth(depth, values)
7.07 µs ± 59.1 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

```

aodntools/timeseries_products/gridded_timeseries.py
diodon: As strange as it may appear, the list approach --convert the np.array to a list and using the depth sort as a list, is a bit faster (5%) than the pure numpy approach. I've used a sample file with ~12K timestamps, which is relatively small compared with the full size file that could have ~100K timestamps.

aodntools/timeseries_products/gridded_timeseries.py
diodon: compression added.

aodntools/timeseries_products/gridded_timeseries.py
ocehugo: as discussed, the speed is secondary here since the difference is not O(1) and the call isn't the bottleneck (which should be all the calls within the for loop - particularly the np.interp one).

aodntools/timeseries_products/gridded_timeseries.py
diodon: Depth values changes in every interaction.

aodntools/timeseries_products/gridded_timeseries.py
diodon: As we discovered, the `groupby` takes a lots of time. So far, constraining the original file to have only de VoI reduces significatively the memory usage and the time. I will explore new ways to do the interpolation

aodntools/timeseries_products/gridded_timeseries.py
diodon: key/values revisited and unified

aodntools/timeseries_products/gridded_timeseries.py
diodon: it is just for removing the first value of all variables that have been set to a dummy, as I cannot concatenate on an empty dataset in the first iteration

ocehugo: Then, just put an 

```
if timestamp == 0:
    Voi_binned = ...
else:
    Voi_binned = xr.concat ...
```



Also, I just noticed that you are using ``` with xr.open_dataset as nc```. I tend to prefer to use ```with``` for short blocks otherwise you affect readability because of the indentation.
If you use just ```nc = xr.open_dataset``` you got a better readability of the code.

ocehugo:
Wouldn't you dump the groupby call? Nonetheless, I can't merge since you need to fix the conflicts.

aodntools/timeseries_products/aggregated_timeseries.py
```diff
@@ -15,7 +15,7 @@
 
 from aodntools import __version__
 
-TEMPLATE_JSON = resource_filename(__name__, 'aggregated_timeseries_template.json')
+TEMPLATE_JSON = resource_filename(__name__, 'gridded_timeseries_template.json')
```
mhidas: Is this an accidental change??
If the template is the same for the two products (or all three??), it should just be called "timeseries_products_template.json".

mhidas: Actually I think it will be easier to keep the templates separate. E.g. the first two products have a range of measured variables, while gridded only has temparature (and possibly salinity). 
So this needs to be changed back to 'aggregated_timeseries_template.json'

aodntools/timeseries_products/Documentation/Gridded_timeseries.md
```diff
@@ -0,0 +1,80 @@
...
+- `site_code`: obtained from the input files (should be the same in all of them);
+- `time_coverage_start`, `time_coverage_end`: set to the full range of TIME values in the aggregated file;
+- `geospatial_vertical_min/max`: set to the full range of DEPTH values in the aggregated file;
+- `geospatial_lat_min/max`: set to the full range of LATITUDE values in the aggregated file;
+- `geospatial_lon_min/max`: set to the full range of LONGITUDE values in the aggregated file;
+- `date_created`: set to the date/time the product file is created;
+- `history`: set to “<date> Hourly aggregated file created. <date>: Gridded file created.”;
+- `included_values_flagged_as`: "Good_data, Probably_good_data"  
+- `keywords`: set to a comma-separated list of the main variable names (“TEMP, DEPTH, HOURLY, GRIDDED”);
+- `lineage`: a statement about how the file was created, including a link to the code used, and any input parameters (except the input files, which are listed in the source_file variable)
```
mhidas: "... listed in the source_file **global attribute**)"

aodntools/timeseries_products/aggregated_timeseries.py
diodon: the templates are different for each product. For the gridded is gridded_timeseries_template

aodntools/timeseries_products/gridded_timeseries.py
mhidas: :+1: 
`.update()` is fine with me, but I think some simple re-formatting (shortening lines and removing extra spaces) would help make this more readable... e.g.
```python
    VoI_interpolated.attrs.update({
        'file_version': global_attribute_dictionary['file_version'],
        'source_file': file_name,
        'time_coverage_start': pd.to_datetime(VoI_interpolated.TIME.values.min()).strftime(timeformat),
￼        # etc...
    })
```

aodntools/timeseries_products/aggregated_timeseries.py
diodon: fixed

aodntools/timeseries_products/README.md
```diff
@@ -5,5 +5,4 @@ This code is used to generate aggregated data products from time-series observat
 Documentation:
 - [Aggregated time series (non-velocity)](Documentation/aggregated_timeseries.md)
 - [Hourly time series (non-velocity)](Documentation/Hourly_timeseries.md)
-
-Please use the [issue tracker](https://github.com/aodn/python-aodntools/issues) for feedback and suggestions related to these products.
```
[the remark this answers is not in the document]
diodon: added!

aodntools/timeseries_products/Documentation/Gridded_timeseries.md
```diff
@@ -0,0 +1,80 @@
...
+
+
+## Scope
+
+The code will provide aggregated files for TEMPERATURE at each IMOS mooring site binned into one hour intervals and linearly interpolated at defined target depths. The aggregation process excludes out-of-water data and records not flagged as “good” or “probably good”. QC flags will not be included. 
+
+
+## Input
+
+The input file is the hourly time series aggregated file for the site.
```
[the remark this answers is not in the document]
diodon: OK changed.

aodntools/timeseries_products/Documentation/Gridded_timeseries.md
```diff
@@ -0,0 +1,80 @@
...
+
+The dimensions of the resulting file  are determined as follows:
+
+- `TIME`:  the 1-hour timestamps from the hourly time series product;
+- `DEPTH`: the target depths where the temperature values have been interpolated. 
+
+
+### Variables
+
+The product will only aggregate Sea Water Temperature `TEMP` when two or more lectures exist at different depths at the same timestamp, and the separation of the sensors are not greater than a maximum distance specified for any particular site.
```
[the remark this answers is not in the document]
diodon: done

aodntools/timeseries_products/Documentation/Gridded_timeseries.md
```diff
@@ -0,0 +1,80 @@
...
+
+- `LATITUDE`: LATITUDE.
+- `LONGITUDE`: LONGITUDE.
+- `TEMP_count`: number of observations in the water column used for the interpolation at every timestamp
+
+### Attributes
+
+The variable attributes will comply with the IMOS metadata standards.
+
+The global metadata will be a set of IMOS standard attributes. Fixed attributes are read from a [JSON file](https://github.com/aodn/python-aodntools/blob/master/aodntools/timeseries_products/hourly_timeseries_template.json) that contains the {key:value} pairs for each of them.
```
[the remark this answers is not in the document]
diodon: fixed

aodntools/timeseries_products/Documentation/Gridded_timeseries.md
```diff
@@ -0,0 +1,80 @@
...
+
+Attributes specific to each aggregated product, are added as follows:
+
+- `site_code`: obtained from the input files (should be the same in all of them);
```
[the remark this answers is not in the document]
diodon: fixed

aodntools/timeseries_products/Documentation/Gridded_timeseries.md
diodon: fixed

aodntools/timeseries_products/gridded_timeseries.py
```diff
@@ -0,0 +1,268 @@
...
```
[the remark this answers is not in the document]
diodon: yes. fixed

aodntools/timeseries_products/gridded_timeseries_config.json
```diff
@@ -0,0 +1,9 @@
+{
+  "site_code": "GBRPPS",
+  "filename": "IMOS_ANMN-QLD_BOSTZ_20090607_GBRPPS_FV02_hourly-timeseries_END-20101101_C-20191106.nc",
```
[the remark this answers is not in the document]
diodon: Both removed. The site_code is being read from the input file

aodntools/timeseries_products/gridded_timeseries.py
```diff
@@ -0,0 +1,268 @@
...
+    VoI_binned = xr.DataArray(np.full((depth_bin_len, 1), np.nan), coords=[depth_bins, [time_0]],
+                              dims=['DEPTH', 'TIME'])
+    VoI_ndepths = xr.DataArray(np.full(1, 0), coords=[[time_0]], dims=['TIME'])
+
+    ## group nc by individual timestamps
+    VoI_grouped = nc.groupby('TIME')
+
+
+    for timestamp, (name, group) in enumerate(VoI_grouped):
+        time = [name]
```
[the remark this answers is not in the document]
diodon: this was a from and old counter in a previous version of the code. removed.

aodntools/timeseries_products/gridded_timeseries.py
```diff
@@ -0,0 +1,268 @@
...
+            VoI_values = list(group[VoI].values)
+            depth = list(group.DEPTH.values)
+            ## sort depths
+            depth, VoI_values = sort_depths(depth, VoI_values)
+
+            ## check for max separation
+            depth_mask = get_depth_mask(depth_bins=depth_bins, depths=depth, max_separation=max_separation)
+            ## do the interpolation
+            # import pdb
+            # pdb.set_trace()
```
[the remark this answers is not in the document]
diodon: fixed

aodntools/timeseries_products/gridded_timeseries.py
```diff
@@ -0,0 +1,268 @@
...
+            VoI_gridded = np.interp(depth_bins, depth, VoI_values, left=np.nan, right=np.nan)
```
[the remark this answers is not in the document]
diodon: Yeah, a bit confusing. Actually,  I need one dataset(VoI_binned) for concatenating the interpolated TEMP nparray (VoI_gridded) another for depths (VoI_ndepths). Both has to be initialised. The final dataset is built on VoI_interpolated. Let me change some names to make it less confusing.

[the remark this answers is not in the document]
diodon: now, temporal values of TEMP (interpolated) and target DEPTH are in the data arrays VoI_temp_tmp and VoI_ndepth_tmp. Both will accumulate in the VoI_temp and VoI_ndepths data arrays and once the loop is over, build the final dataset VoI_interpolated.

aodntools/timeseries_products/gridded_timeseries.py
mhidas: :+1: I agree.
Also, this conversion doesn't seem to have any effect on the final output file, which has TEMP as a `double`. I think a `float` would be sufficient.

PR #91, opened 2020-01-30
Aggregated timeseries quick fix

diodon:
this PR fixes the name of the template JSON file for the aggregated (non-velocity) timeseries


mhidas:
:+1: Thanks!

PR #92, opened 2020-02-02
Velocity aggregated time series product

diodon:
This product flattens UCUR, VCUR, WCUR and reference the values to its TIME and absolute DEPTH. The values are aggregated from all deployments at one site in an indexed ragged array structure with OBSERVATION and INSTRUMENT as the sole dimensions


aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,316 @@
...
+    :param files_to_agg: List of files to sort
+    :return: sorted list of files
+    """
+
+    time_start = []
+    for file in files_to_agg:
+        with nc4.Dataset(file, 'r') as ds:
+            time_start.append(np.datetime64(ds.time_deployment_start))
+    tuples = sorted(zip(time_start, files_to_agg))
+    return [t[1] for t in tuples]
```
ocehugo: 1. change ```files_to_agg``` to ```files``` or file_list or file_str_list -> there is no agg here and put the type intention.
2. Change file to filestr in the loop -> again type intention.
2. ```return [file for _,file in tuples]``` is clearer

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,316 @@
...
+    :param site_code: code of the mooring site
+    :return: dictionary with the file name and list of failed tests
+    """
+
+    attributes = list(nc.attrs)
+    file_variables = list(nc.variables)
+    allowed_dimensions = ['TIME', 'LATITUDE', 'LONGITUDE', 'HEIGHT_ABOVE_SENSOR']
+    required_variables = ['UCUR', 'VCUR', 'WCUR']
+    error_list = []
+
```
ocehugo: Wouldn't the allowed_dimensions and required_variables be better in a global variable (outside the functional scope?)?

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,316 @@
...
+    nc_site_code = nc.site_code
```
ocehugo: I think using "nc.site_code" directly on the if is better - you are not adding any information or scope by naming it again.

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,316 @@
...
+    if nc_site_code != site_code:
+        error_list.append('Wrong site_code: ' + nc_site_code)
+
+    nc_file_version = nc.file_version
```
ocehugo: same as above

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,316 @@
...
+                if dim not in allowed_dimensions:
+                    error_list.append('not allowed dimension: ' + dim)
+
+    if 'NOMINAL_DEPTH' not in file_variables and 'instrument_nominal_depth' not in attributes:
+        error_list.append('no NOMINAL_DEPTH')
+
+    return error_list
+
+
+def get_nvalues(nc):
```
ocehugo: nvalues means nothing here, hence intent is not clear. 

e.g. get_number_of_vertical_cells is much better - I don't even need to read the docstring.

ocehugo: Actually,why not just return the HEIGHT and the TIME variables? The intent will be clearer in the function call - The name would be "get_dim_len" or alike.

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,316 @@
...
+    if 'HEIGHT_ABOVE_SENSOR' in nc.dims:
+        nbins = nc.dims['HEIGHT_ABOVE_SENSOR']
+        nvalues = nc.dims['TIME'] * nbins
+    else:
+        nbins = 1
+        nvalues = nc.dims['TIME']
+    return nvalues, nbins
+
+
+def get_varvalues(nc, varname):
```
ocehugo: ```def flat_variable``` is clearer IMO

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,316 @@
...
+def get_instrumentID(nc):
+    """
+    Create instrument id based on deployment metadata
+    :param nc: xarray dataset
+    :return: instrumentID as string
+    """
+    return '; '.join([nc.deployment_code, nc.instrument, nc.instrument_serial_number])
+
+
+def in_water(nc):
```
ocehugo: You don't need to use the "nc" class as argument, you just need an array and two strings.
Hence, If it defined as ```def get_in_water(time, start_str, end_str)```,
the code call would be:

```
time = nc['TIME']
in_water_ind = get_in_water(time,nc['time_coverage_start'],nc['time_coverage_end'])
```

Again, intent is clear at the function call.

ocehugo: I just noticed that the variable name here is confusing - you use ```nc``` to reference to a ```xarray``` dataset. Just rename the variable to ```xrobj ``` or ```xobj```. I did the comment above expecting the arguments to be netCDF4 objects.

This is a case where type hints shines.

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,316 @@
...
+    """
+    cut data to in-water only timestamps, dropping the out-of-water records.
```
ocehugo: "cut the entire dataset to"

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,316 @@
...
+    Aggregate U, V and W CUR variables from all deployments at one site.
+    the vertical cells are flattened and related to its depth
+    additional metadata variables are stored to track the origin of the data
+    :param files_to_agg: list of files to aggregate
+    :param site_code: site code
+    :param base_path: base path to store the resulting file
+    :return: name of the resulting file, list of rejected files
+    """
+
+    varlist = ['UCUR', 'VCUR', 'WCUR', 'DEPTH']
```
ocehugo: This block of vars can be defined (promoted) to keyword arguments and be dynamically assigned, if required.

```
agg_options = {'varlist': ..., 'time_units': , ...] # or even a global dict
aggregate_velocity(...,varlist=[...],time_units=...). # or **agg_options
```

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,316 @@
...
+    epoch = np.datetime64("1950-01-01T00:00:00")
+    one_day = np.timedelta64(1, 'D')
+
+    varlen_list = []
+    bad_files = []
+    varlen_file = []
+    rejected_files = []
+
+    # default name for temporary file. It will be renamed at the end
+    outfile = 'Velocity_agg_tmp.nc'
```
ocehugo: use random names - if calling more than one thread at the same time things can get ugly

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,316 @@
...
+
+    ## sort the file list in chronological order
+    files_to_agg = sort_files(files_to_agg)
```
ocehugo: sorted_files is better.

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,316 @@
...
+
+    ## check files and get total number of flattened obs
+    for file in files_to_agg:
+        with xr.open_dataset(file) as nc:
+            ## clip to in water data only
+            nc = in_water(nc)
```
ocehugo: in_water_only = in_water(nc)

ocehugo: or alike

ocehugo: dont hide the state mutation. In another words, do not reuse the name if the state changed - There is no copies here only views

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,316 @@
...
+
+            varlen_file.append(get_nvalues(nc))
```
ocehugo: You actually dont need a 15 line get_nvalues function to execute two lines:

```
DIM_X = 1 if 'HEIGHT_ABOVE_SENSOR' not in nc else x['HEIGHT_ABOVE_SENSOR'].size
file_size = DIM_X*nc['TIME'].size
```

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,316 @@
...
+            error_list = check_file(nc, site_code)
```
ocehugo: check first - move this line and the if statements atop

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,316 @@
...
+            if not error_list:
+                varlen_list.append(get_nvalues(nc)[0])
+            else:
+                bad_files.append([file, error_list])
+                rejected_files.append(file)
```
ocehugo: you can create a "good_files' list instead and get rid of the loop in line 175. You can print bad files or just estimate the bad files when reporting ```[x for x in file_agg if x not in good_files]```

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,316 @@
...
+
+    ## remove bad files form the list
+    for file in bad_files:
+        files_to_agg.remove(file[0])
+
+
+    varlen_list = [0] + varlen_list
```
ocehugo: check emptiness instead of doing this.

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,316 @@
...
+    varlen_total = sum(varlen_list)
+    n_files = len(files_to_agg)
+
+
+    ## create ncdf file, dimensions and variables
+    ds = nc4.Dataset(os.path.join(base_path, outfile), 'w')
+    OBSERVATION = ds.createDimension('OBSERVATION', size=varlen_total)
+    INSTRUMENT = ds.createDimension('INSTRUMENT', size=n_files)
+
```
ocehugo: create dictionary with options:

```
obs_float_template = {'datatype':'float','zlib':True,'dimensions':('OBSERVATIONS'),"fill_value":99999.0}
obs_byte_template = {'datatype:'byte','zlib':True,'dimensions':('OBSERVATIONS'),'fill_value':99}
...
vdefs['UCUR'] = {varname:'UCUR',**float_template}
vdefs['UCURqc'] = {varname:'UCURqc',**byte_template};
...
for vname,vopts in vdefs.items():
  vars[k] = ds.createVariable(**vopts)
```

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,316 @@
...
+            end = sum(varlen_list[:index + 2])
+            n_cells = get_nvalues(nc)[1]
+            UCUR[start:end] = get_varvalues(nc, 'UCUR')
+            UCURqc[start:end] = get_varvalues(nc, 'UCUR_quality_control')
+            VCUR[start:end] = get_varvalues(nc, 'VCUR')
+            VCURqc[start:end] = get_varvalues(nc, 'VCUR_quality_control')
+            if 'WCUR' in nc.data_vars:
+                WCUR[start:end] = get_varvalues(nc, 'WCUR')
+                WCURqc[start:end] = get_varvalues(nc, 'WCUR_quality_control')
+            else:
```
ocehugo: There can be a case where WCUR is only found in some files...I would imagine that, if I asked for WCUR in varlist, I would have a partially filled vector in the case of missing variables in some files.

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,316 @@
...
+    for index, file in enumerate(files_to_agg):
+        print(index)
+        with xr.open_dataset(file) as nc:
+
+            ## in water data only
+            nc = in_water(nc)
+
+            start = sum(varlen_list[:index + 1])
+            end = sum(varlen_list[:index + 2])
+            n_cells = get_nvalues(nc)[1]
```
ocehugo: Again - it's not easy to guess what the the second argument returned by get_nvalues is because the function name is not useful (you have to dig into the docstring).

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,316 @@
...
+                WCURqc[start:end] = np.full(varlen_list[index], np.nan)
+            ##calculate depth
+            if 'HEIGHT_ABOVE_SENSOR' in nc.dims:
+                DEPTH[start:end] = (nc.DEPTH + nc.HEIGHT_ABOVE_SENSOR.T).values.flatten()
+                DEPTHqc[start:end] = np.array(n_cells * [nc.DEPTH_quality_control.values]).flatten()
+            else:
+                DEPTH[start:end] = nc.DEPTH.values
+                DEPTHqc[start:end] = nc.DEPTH_quality_control.values
+            ## set TIME and instrument index
+            TIME[start:end] = (np.repeat(get_varvalues(nc, 'TIME'), n_cells) - epoch) / one_day
```
ocehugo: The expression inside the parenthesis is a bit long. Since this is an assignement operation, I would do without a big parenthesis 
```
TIME[start:end] = np.repeat(...)/one_day - epoch/one_day
```
The intent of normalization is much better (and should be slightly quicker).

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,316 @@
...
+            instrument_index[start:end] = np.repeat(index, varlen_list[index + 1])
+            ## get and store deployment metadata
+            LATITUDE[index] = nc.LATITUDE.values
+            LONGITUDE[index] = nc.LONGITUDE.values
+            NOMINAL_DEPTH[index] = TStools.get_nominal_depth(nc)
```
ocehugo: I don't understand the use of this package - TStools. You are only getting netcdf attributes. do you really need to import this package to do that?

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,316 @@
...
+        attribute_dictionary = json.load(json_file)
+    variable_attribute_dictionary = attribute_dictionary['_variables']
+    global_attribute_dictionary = attribute_dictionary['_global']
+
+    ## set variable attrs
+    for var in list(ds.variables):
+        ds[var].setncatts(variable_attribute_dictionary[var])
+
+    ## set global attrs
+    timeformat = '%Y-%m-%dT%H:%M:%SZ'
```
ocehugo: another two options - should be kwargs.

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,316 @@
...
+    file_timeformat = '%Y%m%d'
+
+    time_start = nc4.num2date(np.min(TIME[:]), time_units, time_calendar).strftime(timeformat)
```
ocehugo: call this block a function 
```
set_global_attrs(ds,time_units,time_calendar,time_format,...
options={...}
set_global_attrs(ds,**options)
```

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,316 @@
...
+                    'contributor_email':       "; ".join(contributor_email),
+                    'contributor_role':        "; ".join(contributor_role)}
+
+    global_attribute_dictionary.update(add_attribute)
+    ds.setncatts(dict(sorted(global_attribute_dictionary.items())))
+
+    ds.close()
+
+
+    ## create the output file name and rename the tmp file
```
ocehugo: another case for a function ``` output_name = rename_file(oldname,newname) ```.
Also, you need to check for overwrites (rename to an already existing one).

Personally, I don't think you need to use "tmp" files or to rename at all - just do the kind of checks you are doing here and above before processing/concat the files (reading attributes, etc). If you think on this steps (check stuff, create a valid name) as a validation step, then you only spend time in valid inputs.

For example, What is the use of an entire concatenated array in memory if you cant find the site_code, facility_code, contributors, etc? Check the easy things first...

aodntools/timeseries_products/velocity_aggregated_timeseries_template.json
```diff
@@ -0,0 +1,169 @@
+{
```
ocehugo: I imagine you probably have a lot of templates at the moment - maybe a time for refactoring?

For example, a lot of fields are probably repeats from other templates - maybe a template to rule them all, while small templates only put stuff to "overwrite" bits. This may also avoid spread typos, slightly different entries,  and help in maintenance (imagine something need to change - you have to remember and change all templates...).

ocehugo:
@diodon, don't forget to fix the aggregate_timeseries part - tests are not passing and there is some warnings about duplicated fill_values in the ncwriter.

aodntools/timeseries_products/velocity_aggregated_timeseries.py
diodon: ok

aodntools/timeseries_products/velocity_aggregated_timeseries.py
diodon: changed

aodntools/timeseries_products/velocity_aggregated_timeseries.py
diodon: I've renamed the function and variables for clarity. The objective is to return the number of values that result in the flattening of the grid

aodntools/timeseries_products/velocity_aggregated_timeseries.py
diodon: Agree, changed

aodntools/timeseries_products/velocity_aggregated_timeseries.py
diodon: added

aodntools/timeseries_products/velocity_aggregated_timeseries.py
diodon: changed to UUID name

aodntools/timeseries_products/velocity_aggregated_timeseries.py
diodon: changed

aodntools/timeseries_products/velocity_aggregated_timeseries.py
diodon: solved

aodntools/timeseries_products/velocity_aggregated_timeseries_template.json
diodon: it is in the plan

aodntools/timeseries_products/velocity_aggregated_timeseries.py
diodon: the package contains several utility functions. I use three of them in the code. In the future a refactored version of it will be a package with functions used by all product-generating scripts.

aodntools/timeseries_products/velocity_aggregated_timeseries.py
diodon: changed

aodntools/timeseries_products/velocity_aggregated_timeseries.py
diodon: nc for xarray and ds to netcdf4 dataset

aodntools/timeseries_products/velocity_aggregated_timeseries.py
diodon: could be but we don't expect to aggregate anything else apart from those variables.

aodntools/timeseries_products/velocity_aggregated_timeseries.py
diodon: zero is needed as the start index in the main loop.

aodntools/timeseries_products/velocity_aggregated_timeseries.py
ocehugo: see comment below about this function - it's not necessary.

aodntools/timeseries_products/velocity_aggregated_timeseries.py
ocehugo: After going down and back again I understand what this variable is.Hence, uou should rename it -> It;s not a varlen_list, it's an index list.

Also, you may pre-compute the start/end of the indexes (line 214/215) here instead and use it in the for loop (also easier to debug the index ranges).

aodntools/timeseries_products/velocity_aggregated_timeseries.py
ocehugo: IMO it's not required -> it's just noise for very little gain

aodntools/timeseries_products/velocity_aggregated_timeseries.py
ocehugo: anyway, this is a good practice both codewise and intent wise. For example, when reading, why the function "aggregate velocity" does not provide the option to select which velocity/vars I want to aggregate? 

Code wise, "default" options should hardly be defined in the scope of the function. Also, if you put as kwargs and need to change the parameters you just change the function call instead of changing the function code!

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,316 @@
...
+            VCURqc[start:end] = get_varvalues(nc, 'VCUR_quality_control')
+            if 'WCUR' in nc.data_vars:
+                WCUR[start:end] = get_varvalues(nc, 'WCUR')
+                WCURqc[start:end] = get_varvalues(nc, 'WCUR_quality_control')
+            else:
+                WCUR[start:end] = np.full(varlen_list[index], np.nan)
+                WCURqc[start:end] = np.full(varlen_list[index], np.nan)
+            ##calculate depth
+            if 'HEIGHT_ABOVE_SENSOR' in nc.dims:
+                DEPTH[start:end] = (nc.DEPTH + nc.HEIGHT_ABOVE_SENSOR.T).values.flatten()
```
mhidas: `HEIGHT_ABOVE_SENSOR` is in the opposite direction to `DEPTH`, so it should be `DEPTH - HEIGHT_ABOVE_SENSOR`!

Also, I don't quite understand the use of the `.T` attribute. I'm assuming it's meant to be a transpose, but `HEIGHT_ABOVE_SENSOR` is a 1-dimensional array. ([Xarray](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.html#xarray.DataArray) doesn't seem to provide any documentation of this attribute, while [numpy](https://numpy.org/doc/1.16/reference/generated/numpy.ndarray.T.html#numpy.ndarray.T) says it just return the original array if ndim < 2)

diodon: Yes, you're correct. The .T has no effect as HEIGHT_ABOVE_SENSOR is 1D.fixed

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -225,7 +225,7 @@ def aggregate_velocity(files_to_agg, site_code, base_path):
                 WCURqc[start:end] = np.full(varlen_list[index], np.nan)
             ##calculate depth
             if 'HEIGHT_ABOVE_SENSOR' in nc.dims:
-                DEPTH[start:end] = (nc.DEPTH + nc.HEIGHT_ABOVE_SENSOR.T).values.flatten()
```
ocehugo: Since this was undetected last time, I think a validation would be worth here (Just in case).

ocehugo:
IMO: The most important structural changes were not applied :)

1. The fact you need temporary files.
    b. Which is caused by a lack of earlier validation.
    c. Which also leads to the fact that, if there is anything wrong in the file, you will have dangling temporary files.
    d. Which will lead to cleanup callbacks/cases, calls to "rm", etc.
2. The WCUR missing values.
3. The way of global and custom options are set in the code (fixed options are not "globals", runtime options are not kwargs).
    a. Which may cause hard to test when code is changed, some new code to be incompatible or cases not being handle, etc.

Finally, I don't see the tests. For example, the HEIGHT above sensor raise issue could've been easily detected.
Others:
1. Test metadata
2. Test DEPTH is correct/expected.
3. No data where DEPTH is not defined.
4. Data is valid and compares to original/unaggregated files.
etc

aodntools/timeseries_products/velocity_aggregated_timeseries.py
diodon: check added to check_file function

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,373 @@
...
+                    'rejected_files':           "\n".join(rejected_files),
+                    'contributor_name':        "; ".join(contributor_name),
+                    'contributor_email':       "; ".join(contributor_email),
+                    'contributor_role':        "; ".join(contributor_role),
+                    'generating_code_version':  __version__
+    }
+
+    ## add version
+    github_comment = ('\nThis file was created using https://github.com/aodn/python-aodntools/blob/'
+                      '{v}/aodntools/timeseries_products/aggregated_timeseries.py'.format(v=__version__)
```
mhidas: Forgot to update the file name! Actually, we could do this instead...
```suggestion
                      '{v}/aodntools/timeseries_products/{f}'.format(v=__version__, f=os.path.basename(__file__))
```

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,373 @@
...
+
+    return ncout_path, bad_files
+
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser(description="Concatenate X,Y,Z velocity variables from ALL instruments from ALL deployments from ONE site")
+    parser.add_argument('-site', dest='site_code', help='site code, like NRMMAI',  required=True)
+    parser.add_argument('-files', dest='filenames', help='name of the file that contains the source URLs', required=True)
+    parser.add_argument('-path', dest='output_path', help='path where the result file will be written. Default: ./', default='./', required=False)
```
mhidas: `output_path` is no longer used, so you can remove this line.

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,373 @@
...
+            UCUR[start:end] = flat_variable(nc, 'UCUR')
+            UCURqc[start:end] = flat_variable(nc, 'UCUR_quality_control')
+            VCUR[start:end] = flat_variable(nc, 'VCUR')
+            VCURqc[start:end] = flat_variable(nc, 'VCUR_quality_control')
+            if 'WCUR' in nc.data_vars:
+                WCUR[start:end] = flat_variable(nc, 'WCUR')
+                WCURqc[start:end] = flat_variable(nc, 'WCUR_quality_control')
+            else:
+                WCUR[start:end] = np.full(varlen_list[index], np.nan)
+                WCURqc[start:end] = np.full(varlen_list[index], 9)
```
mhidas: As mentioned in my comment on the `check_file` function, if we reject files that don't have `WCUR`, then this case isn't needed here.
But if it _is_ needed, I think the array length here is `varlen_list[index+1]` (as for `instrument_index`, below)

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,373 @@
...
+            ##calculate depth
+            if 'HEIGHT_ABOVE_SENSOR' in nc.dims:
+                DEPTH[start:end] = (nc.DEPTH - nc.HEIGHT_ABOVE_SENSOR).values.flatten()
+                DEPTHqc[start:end] = np.array(n_cells * [nc.DEPTH_quality_control.values]).flatten()
```
mhidas: Looks a bit simpler and consistent with how you did the same operation for `TIME` and `instrument_index` below
```suggestion
                DEPTHqc[start:end] = np.repeat(nc.DEPTH_quality_control, n_cells).values
```

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,373 @@
...
+import numpy as np
+import json
+from datetime import datetime
+import argparse
+from pkg_resources import resource_filename
+from aodntools import __version__
+
+import xarray as xr
+
+import aggregated_timeseries as TStools
```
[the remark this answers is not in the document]
diodon: OK.

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,373 @@
...
+
+TEMPLATE_JSON = resource_filename(__name__, 'velocity_aggregated_timeseries_template.json')
+
+
+def sort_files(file_list, input_dir=""):
```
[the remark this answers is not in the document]
diodon: deleted from the source, imported from tools

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,373 @@
...
+
+
+def check_file(nc, site_code):
+    """
+    Return list of errors found in the file:
+    Variables of interest are present
+    TIME, DEPTH, LATITUDE, LONGITUDE,  is present
+    NOMINAL_DEPTH is not present as variable or attribute
+    file_version is not FV01
+    if LATITUDE or LONIGITUDE dimension has length >1
```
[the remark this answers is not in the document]
diodon: fixed!

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,373 @@
...
+
+    :param nc: xarray dataset
+    :param site_code: code of the mooring site
+    :return: dictionary with the file name and list of failed tests
+    """
+
+    attributes = list(nc.attrs)
+    variables = list(nc.variables)
+    allowed_dimensions = ['TIME', 'LATITUDE', 'LONGITUDE', 'HEIGHT_ABOVE_SENSOR']
+    required_variables = ['UCUR', 'VCUR', 'WCUR']
```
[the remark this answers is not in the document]
diodon: I've removed the WCUR check from the file_check function. I think we can leave the WCUR handle in the code even if WCUR are expected to be present. It will be safe in case the variable disappears in the future. It does not harm.

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,373 @@
...
+        error_list.append('Wrong site_code: ' + nc.site_code)
+
+    nc_file_version = nc.file_version
+    if 'Level 1' not in nc_file_version:
+        error_list.append('Wrong file version: ' + nc_file_version)
+
+    if 'DEPTH' not in variables:
+        error_list.append('DEPTH variable missing')
+
+    if 'DEPTH' not in variables and 'HEIGHT_ABOVE_SENSOR' not in variables:
```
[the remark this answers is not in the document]
diodon: OK

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,373 @@
...
+    """
+    Return a 1D array of 2D values
+    :param nc: dataset
+    :param varname: Variable of interest
+    :return: variable values flattened
+    """
+    return nc[varname].values.flatten()
+
+
+def get_instrument_id(nc):
```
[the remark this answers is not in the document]
diodon: removed

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,373 @@
...
+    """
+
+    varlist = ['UCUR', 'VCUR', 'WCUR', 'DEPTH']
+    time_units="days since 1950-01-01 00:00:00 UTC"
+    time_calendar="gregorian"
+    epoch = np.datetime64("1950-01-01T00:00:00")
+    one_day = np.timedelta64(1, 'D')
+
+    varlen_list = []
+    bad_files = []
```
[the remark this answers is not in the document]
diodon: sweet! changed.

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,373 @@
...
+    ## sort the file list in chronological order
+    files_to_agg = sort_files(files_to_agg, input_dir=input_dir)
+
+    ## check files and get total number of flattened obs
+    for file in files_to_agg:
+        with xr.open_dataset(os.path.join(input_dir, file)) as nc:
+            ## clip to in water data only
+            nc = in_water(nc)
+
+            varlen_file.append(get_number_flatvalues(nc))
```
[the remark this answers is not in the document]
diodon: in fact, when drop=True is used, it returns a new object. The first intention is not to work with masked arrays in the aggregation. We may want to include the drop argument in the in_water function if we want to have a view using drop=False.

the call is now done with good files only

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,373 @@
...
+            error_list = check_file(nc, site_code)
+            if not error_list:
+                varlen_list.append(get_number_flatvalues(nc)[0])
+            else:
+                bad_files.append([file, error_list])
+                rejected_files.append(file)
+
+    ## remove bad files form the list
+    for file in bad_files:
+        files_to_agg.remove(file[0])
```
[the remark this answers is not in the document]
diodon: yes, fixed.

aodntools/timeseries_products/velocity_aggregated_timeseries.py
diodon: removed

aodntools/timeseries_products/velocity_aggregated_timeseries.py
diodon: That is correct!

aodntools/timeseries_products/velocity_aggregated_timeseries.py
diodon: yes indeed

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,317 @@
...
+import numpy as np
+import json
+from datetime import datetime
+import argparse
+from pkg_resources import resource_filename
+from aodntools import __version__
+
+import xarray as xr
+
+import aggregated_timeseries as utils
```
mhidas: This may work when you run the code in your terminal, but it won't work in general. Please specify the full path within the package:
```suggestion
from aodntools.timeseries_products import aggregated_timeseries as utils
```

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -0,0 +1,317 @@
...
+    Variables of interest are not  present
+    TIME, DEPTH, LATITUDE, LONGITUDE, are not present
+    NOMINAL_DEPTH is not present as variable or attribute
+    file_version is not FV01
+    the variable has one or more dimension not in allowed dimensions
+    if LATITUDE or LONGITUDE dimension has length >1
+
+    :param nc: xarray dataset
+    :param site_code: code of the mooring site
+    :return: dictionary with the file name and list of failed tests
```
mhidas:
```suggestion
    :return: list of failed tests
```

aodntools/timeseries_products/velocity_aggregated_timeseries.py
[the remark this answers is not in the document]
mhidas: :+1:

aodntools/timeseries_products/Documentation/velocity_aggregated_timeseries.md
```diff
@@ -0,0 +1,124 @@
...
+
+The code is able to access the input files either locally, or remotely via the OPeNDAP protocol. 
+
+## Method
+
+Generating function: 
+
+```
+usage: velocity_aggregated_timeseries.py [-h] -site SITE_CODE -files FILENAMES
+                                         [-path OUTPUT_PATH]
```
mhidas: Remove `-path`, and also below.

aodntools/timeseries_products/Documentation/velocity_aggregated_timeseries.md
```diff
@@ -0,0 +1,124 @@
...
+
+Before proceeding to the aggregation, each input file will be checked to ensure it meets the requirements (as specified above under Inputs). Any input files that fail to meet the requirements will be excluded from the aggregation, and their URL listed in a global attribute `rejected_files`.
+
+### Dimensions
+
+The dimensions of the resulting file  are determined as follows:
+
+- `OBSERVATION`:    the total number of observation records, excluding out-of-the-water data, in all input files;
+- `INSTRUMENT`:     the number of instruments (i.e. number of files);
+- `string`:         a string dimension for character array variables.
```
mhidas: `string` is not needed (unless we change back to the char array format)

aodntools/timeseries_products/Documentation/velocity_aggregated_timeseries.md
```diff
@@ -0,0 +1,124 @@
...
+
+The `DEPTH` variable from input files is concatenated into a variable `DEPTH(OBSERVATION)`. In the case of ADCP instruments, the `HEIGH_ABOVE_SENSOR`  is converted to absolute depth by subtracting each of the height values from the depth measurements at the instrument. `DEPTH_quality_control`, if present, is also included. 
+
+All output variables with the `INSTRUMENT` dimension are sorted in chronological order, and the input files aggregated chronologically, according to the global attribute time_deployment_start.
+
+In order to keep track of the provenance of the aggregated file, accessory variables are created:
+
+- `instrument_index(OBSERVATION)`: index [0:number of files] of the instrument used, referencing the `INSTRUMENT` dimension.
+- `source_file(INSTRUMENT, string)`: URLs of the files used
+- `instrument_id(INSTRUMENT, string)`: concatenated deployment_code, instrument and instrument_serial_number from the global attributes of each file
```
mhidas: Again, remove `string` dimension here, unless we end up including it.

aodntools/timeseries_products/Documentation/velocity_aggregated_timeseries.md
```diff
@@ -0,0 +1,124 @@
...
+- `LONGITUDE(INSTRUMENT)`: LONGITUDE per instrument.
+- `NOMINAL_DEPTH(INSTRUMENT)`: nominal depth per instrument, from the input file’s variable `NOMINAL_DEPTH` or global attribute instrument_nominal_depth.
+- `SECONDS_TO_MIDDLE(INSTRUMENT)`:  offset from the timestamp to the middle of the measurement window for each deployment
+
+
+### Attributes
+
+The variable attributes will comply with the IMOS metadata standards.
+
+The global metadata will be a set of IMOS standard attributes. Fixed attributes are read from a JSON file that contains the {key:value} pairs for each of them. See the contents of this file at the end of this document.
```
mhidas:
```suggestion
The global metadata will be a set of IMOS standard attributes. Fixed attributes are read from a [JSON file](aodntools/timeseries_products/velocity_aggregated_timeseries_template.json) that contains the {key:value} pairs for each of them.
```

aodntools/timeseries_products/Documentation/velocity_aggregated_timeseries.md
```diff
@@ -0,0 +1,124 @@
...
+
+- `site_code`: obtained from the input files (should be the same in all of them);
+- `time_coverage_start`, `time_coverage_end`: set to the full range of TIME values in the aggregated file;
+- `geospatial_vertical_min`, `geospatial_vertical_max`: set to the full range of DEPTH values in the aggregated file;
+- `geospatial_lat_min`, `geospatial_lat_max` : set to the full range of LATITUDE values in the aggregated file;
+- `geospatial_lon_min`, `geospatial_lon_max`: set to the full range of LONGITUDE values in the aggregated file;
+- `date_created`: set to the date/time the product file is created;
+- `history`: set to “<date_created>: Aggregated file created.”;
+- `keywords`: set to a comma-separated list of the main variable names (“UCUR, VCUR, WCUR, DEPTH, AGGREGATED”);
+- `lineage`: a statement about how the file was created, including a link to the code used, and any input parameters (except the input files, which are listed in the source_file variable)
```
mhidas: We're not actually including the input parameters, so we may as well not say that here.

aodntools/timeseries_products/Documentation/velocity_aggregated_timeseries.md
```diff
@@ -0,0 +1,124 @@
...
+- `title`: "Long Timeseries Velocity Aggregated product: UCUR, VCUR, WCUR, DEPTH at <site_code>  between <time_coverage_start> and <time_coverage_end>"
+- `rejected_files`: a list of URLs for files that were in the input files list, but did not meet the input requirements. 
+
+
+## Output
+
+The output from a single run of the code will be an aggregated file of all available measurements of a single non-velocity variable at one mooring site.
```
mhidas:
```suggestion
The output from a single run of the code will be an aggregated file of all available current velocity measurements at one mooring site.
```

diodon: changed

aodntools/timeseries_products/Documentation/velocity_aggregated_timeseries.md
diodon: changed

aodntools/timeseries_products/Documentation/velocity_aggregated_timeseries.md
diodon: changed

aodntools/timeseries_products/Documentation/velocity_aggregated_timeseries.md
```diff
@@ -0,0 +1,124 @@
...
+- [Input](#input)
+- [Method](#method)
+- [Output](#output)
+
+
+
+
+## Objective
+
+This product provides aggregated U, V, and W velocity time-series files for each mooring site, without any interpolation or filtering, except for the exclusion of the out-of-water data. For the ADCP instruments, the absulte depth of the measuring cell is calculated using the `DEPTH` measured at the instrument and the `HEIGHT_ABOVE_SENSOR`, The Quality Control (QC) flags are preserved. All the (python) code used for the generation of the products is openly available on GitHub.
```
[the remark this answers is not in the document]
diodon: changed

aodntools/timeseries_products/Documentation/velocity_aggregated_timeseries.md
diodon: string removed

aodntools/timeseries_products/Documentation/velocity_aggregated_timeseries.md
diodon: removed

aodntools/timeseries_products/Documentation/velocity_aggregated_timeseries.md
diodon: updated to the new arguments

aodntools/timeseries_products/velocity_aggregated_timeseries.py
diodon: changed

aodntools/timeseries_products/velocity_aggregated_timeseries.py
diodon: changed

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -200,13 +203,18 @@ def velocity_aggregated(files_to_agg, site_code, input_dir='', output_dir='./',
             else:
                 WCUR[start:end] = np.full(varlen_list[index + 1], np.nan)
                 WCURqc[start:end] = np.full(varlen_list[index + 1], 9)
-            ##calculate depth
+
+            ##calculate depth and add CELL_INDEX
             if 'HEIGHT_ABOVE_SENSOR' in nc.dims:
                 DEPTH[start:end] = (nc.DEPTH - nc.HEIGHT_ABOVE_SENSOR).values.flatten()
                 DEPTHqc[start:end] = np.repeat(nc.DEPTH_quality_control, n_cells).values
+                CELL_INDEX[start:end] = np.array([np.full(n_measurements, cell) for cell in range(n_cells)],
```
mhidas: This gives you the incorrect order of values. The output variables cycle through all the cells for a given timestamp, then the next timestamp, etc... So I think what you want is
```suggestion
                CELL_INDEX[start:end] = np.tile(np.arange(n_cells, dtype=np.uint32))
```

mhidas: (oops, left out the `n_measurements` arg in my initial suggestion - added now)

mhidas:
You applied my documentation suggestions to the wrong file (aggregated_timeseries.md instead of velocity_aggregated_timeseries.md)!

aodntools/timeseries_products/velocity_aggregated_timeseries.py
diodon: Yes. That is terribly wrong. I fixed and now it works just fine.

aodntools/timeseries_products/velocity_aggregated_timeseries_template.json
```diff
@@ -0,0 +1,176 @@
...
+            "flag_values": "0, 1, 2, 3, 4, 5, 6, 7, 8, 9",
+            "long_name": "quality flag for upward_sea_water_velocity",
+            "standard_name": "upward_sea_water_velocity status_flag"
+        },
+        "SECONDS_TO_MIDDLE": {
+            "long_name": "number of seconds to middle of the measurement window"
+        },
+        "CELL_INDEX": {
+            "long_name": "index of the corresponding measuring cell",
+            "comment": "Cell index is included for reference only and cannot be used to extract values at constant depth. The number and vertical spacing of cells can vary by instrument and deployment. The vertical spacing also varies with time during a deployment. The closest cell to the sensor has index 0."
```
mhidas: @diodon @ggalibert Do you think this is clear enough?

mhidas:
Ok, I think this is good enough for a first go!

PR #105, opened 2020-03-05
Velocity aggregated timeseries

diodon:
This PR fixes the problem of DEPTH_quality_control with more than one dimension. Tested in all sites, no problems!


mhidas:
:+1:

PR #78, opened 2019-11-06
Aggregated timeseries metadata updates

mhidas:
Just updating a couple of attributes to better keep track of provenance:
* Make `source_file` entries (optionally) relative to a base path;
* (Optionally) add `source_file` variable attributes to document how each source file can be downloaded or accessed via OPENDAP;
* Add a GitHub URL to version of the code used to generate the file in the `lineage` global attribute;
* Add a separate `generating_code_version` global attribute.

PR #106, opened 2020-03-05
Gridded product metadata attributes

mhidas:
Add code version, code github URL, and source file download/OPENDAP URLs in global attributes of the gridded product. (as done in #78 and #96 for the other two products).

Also limited the global attributes copied from the input file to an specified list (rather than everything).


diodon:
all good!

PR #111, opened 2020-03-17
Replacing deprecated setup.py test entry point with pytest

lwgordonimos:
This brings the test entry point into line with python-aodndata.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,12 @@ python:
   - "3.6"
 
 install:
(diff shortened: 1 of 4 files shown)


lwgordonimos:
For context, I did originally have this rolled up with https://github.com/aodn/python-aodntools/pull/110, but it wasn't required in the end. It is however a useful standardisation, so I put it up separately anyway.

mhidas:
:+1: Thanks @lwgordonimos

PR #119, opened 2020-04-01
Add constraints.txt

lwgordonimos:
Fixes https://github.com/aodn/issues/issues/692


lwgordonimos:
@mhidas yeah, good idea. If the constraint is purely because of Python 3.5, should be moved to constraints.txt.

mhidas:
@lwgordonimos see last commit. Unless I've done something silly, this is good to merge.

PR #120, opened 2020-04-23
feat(fixes): shebangs and cmd-line standards

ocehugo:
Just small annoying bits - missing shebangs and non-standard cmd line arguments lines.
@mhidas, worth checking if this is safe since I didn't check how you handle the calls to these tools .

### DIFF
--- a/aodntools/timeseries_products/aggregated_timeseries.py
+++ b/aodntools/timeseries_products/aggregated_timeseries.py
@@ -422,17 +422,17 @@ def main_aggregator(files_to_agg, var_to_agg, site_code, input_dir='', output_di
 if __name__ == "__main__":
 
     parser = argparse.ArgumentParser(description="Aggregate ONE variable from ALL instruments from ALL deployments from ONE site")
-    parser.add_argument('-site', dest='site_code', help='site code, like NRMMAI',  required=True)
-    parser.add_argument('-var', dest='varname', help='variable to aggregate, like TEMP', required=True)
-    parser.add_argument('-files', dest='filenames',
+    parser.add_argument('--site', dest='site_code', help='site code, like NRMMAI',  required=True)
+    parser.add_argument('--var', dest='varname', help='variable to aggregate, like TEMP', required=True)
+    parser.add_argument('--files', dest='filenames',
                         help='name of the file that contains the source URLs (relative to inpath, if given)',
                         required=True)
-    parser.add_argument('-indir', dest='input_dir', help='base path of input files', default='', required=False)
-    parser.add_argument('-outdir', dest='output_dir', help='path where the result file will be written. Default ./',
+    parser.add_argument('--indir', dest='input_dir', help='base path of input files', default='', required=False)
+    parser.add_argument('--outdir', dest='output_dir', help='path where the result file will be written. Default ./',
                         default='./', required=False)
-    parser.add_argument('-download_url', dest='download_url', help='path to the download_url_prefix',
+    parser.add_argument('--download_url', dest='download_url', help='path to the download_url_prefix',
                         default='', required=False)
-    parser.add_argument('-opendap_url', dest='opendap_url', help='path to the opendap_url_prefix',
+    parser.add_argument('--opendap_url', dest='opendap_url', help='path to the opendap_url_prefix',
                         default='', required=False)
     args = parser.parse_args()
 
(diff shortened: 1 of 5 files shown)


aodntools/timeseries_products/aggregated_timeseries.py
```diff
@@ -422,17 +422,17 @@ def main_aggregator(files_to_agg, var_to_agg, site_code, input_dir='', output_di
 if __name__ == "__main__":
 
     parser = argparse.ArgumentParser(description="Aggregate ONE variable from ALL instruments from ALL deployments from ONE site")
-    parser.add_argument('-site', dest='site_code', help='site code, like NRMMAI',  required=True)
-    parser.add_argument('-var', dest='varname', help='variable to aggregate, like TEMP', required=True)
-    parser.add_argument('-files', dest='filenames',
+    parser.add_argument('--site', dest='site_code', help='site code, like NRMMAI',  required=True)
+    parser.add_argument('--var', dest='varname', help='variable to aggregate, like TEMP', required=True)
+    parser.add_argument('--files', dest='filenames',
```
mhidas, line 427: While we're at it, we could simplify both the code and the command line options by:
1) Turning `required` args into positional ones (so don't need `required=True`);
1) Making the argument names consistent with the destination names used within the code; and
1) Then remove the unnecessary `dest` argument (set from the long option by default - see [docs](https://docs.python.org/3.5/library/argparse.html#dest)).

So e.g. this would become
```suggestion
    parser.add_argument('site_code', help='site code, like NRMMAI')
    parser.add_argument('varname', help='variable to aggregate, like TEMP')
    parser.add_argument('filenames',
                        help='name of the file that contains the source URLs (relative to inpath, if given)')
```

aodntools/timeseries_products/aggregated_timeseries.py
```diff
@@ -422,17 +422,17 @@ def main_aggregator(files_to_agg, var_to_agg, site_code, input_dir='', output_di
...
                         help='name of the file that contains the source URLs (relative to inpath, if given)',
                         required=True)
-    parser.add_argument('-indir', dest='input_dir', help='base path of input files', default='', required=False)
-    parser.add_argument('-outdir', dest='output_dir', help='path where the result file will be written. Default ./',
+    parser.add_argument('--indir', dest='input_dir', help='base path of input files', default='', required=False)
+    parser.add_argument('--outdir', dest='output_dir', help='path where the result file will be written. Default ./',
```
mhidas, line 431: Similarly, for optional arguments. We can also add short options, and `required=False` is [unnecessary as it's the default for an "option" anyway](https://docs.python.org/3.5/library/argparse.html#required))
```suggestion
    parser.add_argument('-i', '--input_dir', help='base path of input files', default='')
    parser.add_argument('-o', '--output_dir', help='path where the result file will be written. Default ./',
                        default='./')
```

aodntools/timeseries_products/aggregated_timeseries.py
```diff
@@ -422,17 +422,17 @@ def main_aggregator(files_to_agg, var_to_agg, site_code, input_dir='', output_di
...
                         default='./', required=False)
-    parser.add_argument('-download_url', dest='download_url', help='path to the download_url_prefix',
+    parser.add_argument('--download_url', dest='download_url', help='path to the download_url_prefix',
```
mhidas, line 433: For these last two it may not be worth adding a short option, but we can still simplify...
```suggestion
    parser.add_argument('--download_url', help='path to the download_url_prefix', default='')
    parser.add_argument('--opendap_url', help='path to the opendap_url_prefix', default='')
```

mhidas:
Note that this changes many of the command-line arguments
- The required parameters have become positional args.
- Some of the optional ones have been renamed to match corresponding variable/arg names within the code.
- Some single-letter alternatives have been added.

This brings the interface more in line with standard command-line tools (on Linux).

mhidas:
@alexgleith Why did you close this?
I still think it's relevant, though not high priority for us as we run this code with a Python process, not from a shell. However, as this is an open-source repo (I know at least one group at UNSW using it), the command-line functionality is there and this changes it in a non-backwards-compatible way, I'm saving this until we're ready to release a v2.

alexgleith:
I closed it because it hadn't been touched for over a year. Sorry :-)

PR #128, opened 2020-07-01
Check QC flag conventions in timeseries products

### LINKED ISSUE #126: Product codes assume IMOS QC flag scheme but don't check for it
The the aggregated product includes QC flags in the output file, while the hourly product excludes data values not flagged as "good" or "prodbably good". Both of these make the implicit assumption that the input file uses the "IMOS standard flags".

More specifically,
* All `_quality_control` variables in the _aggregated product output files_ are given the attributes:
```
"flag_meanings": "No_QC_performed Good_data Probably_good_data Bad_data_that_are_potentially_correctable Bad_data Value_changed Not_used Not_used Not_used Missing_value",
"flag_values": "0, 1, 2, 3, 4, 5, 6, 7, 8, 9",
```
* In the hourly codes, values are only included if their QC flag is 1 or 2 (without checking the "flag_meanings" attribute in each input file).

The assumption is correct for the vast majority of IMOS moorings data, except for the Acidification Moorings, which use the  WOCE QC flags:
```
flag_values = 2b, 3b, 4b ;
flag_meanings = "good questionable bad" ;
```
Luckily this is sort of compatible, in that selecting only flags "1 or 2" gets us only the good data. However, for the aggregated product, re-labelling these flags as if they were from the IMOS scheme is misleading. In effect all the "good" data (flag 2) will now look like it's only "probably_good_data".

mhidas:
Fixes #126


mhidas:
@ocehugo Sorry I forgot to add a test file to the repo. I just added it and force pushed. Should pass the checks now.

aodntools/timeseries_products/common.py
```diff
@@ -6,66 +6,70 @@ class NoInputFilesError(Exception):
...
+    :param variables_of_interest: variable name or list of names to be included in the product
     :return: list of failed tests
     """
 
+    if isinstance(variables_of_interest, str):
+        variables_of_interest = [variables_of_interest]
     attributes = list(nc.attrs)
     variables = list(nc.variables)
-    allowed_dimensions = ['TIME', 'LATITUDE', 'LONGITUDE']
+    allowed_dimensions = {'TIME', 'LATITUDE', 'LONGITUDE'}
```
ocehugo: I don't understand these definitions here. If you wish to maintain them at the function level, put them close to their use. If they are constants that are used elsewhere, define them as constants outside for re-use.

aodntools/timeseries_products/common.py
```diff
@@ -6,66 +6,70 @@ class NoInputFilesError(Exception):
...
         error_list.append('Wrong site_code: ' + nc.site_code)
 
-    nc_file_version = nc.file_version
+    nc_file_version = nc.attrs.get('file_version', '[missing]')
     if 'Level 1' not in nc_file_version:
         error_list.append('Wrong file version: ' + nc_file_version)
 
-    if 'TIME' not in variables:
-        error_list.append('TIME variable missing')
+    for coord in required_coordinates:
```
ocehugo: a set comparison is shorter: ``` required_coordinates.issubset(set(variables))```

aodntools/timeseries_products/velocity_aggregated_timeseries.py
```diff
@@ -137,7 +75,7 @@ def velocity_aggregated(files_to_agg, site_code, input_dir='', output_dir='./',
     n_obs_total = 0
     for file in files_to_agg:
         with xr.open_dataset(os.path.join(input_dir, file)) as nc:
-            error_list = check_file(nc, site_code)
+            error_list = check_file(nc, site_code, 'velocity')
```
ocehugo, line 78: I don't understand the 'velocity' argument, since it's not a variable.

aodntools/timeseries_products/common.py
```diff
@@ -6,6 +6,50 @@ class NoInputFilesError(Exception):
...
+            errors.add('variable {var} not in file'.format(var=var))
+            continue
+        conventions = getattr(nc[var], 'quality_control_conventions', None)
+        if conventions is None:
+            errors.add('variable {var} missing quality_control_conventions'.format(var=var))
+            continue
+        if conventions not in accepted_conventions:
+            errors.add('unexpected quality_control_conventions: "{conventions}"'.format(conventions=conventions))
+
+    return sorted(errors)
```
ocehugo, line 50: Maybe is better to sort the varnames instead and use a list for errors, as is done elsewhere.

ocehugo:
Two things made harder to review this bit:

1. Somehow github is not playing well with the force push. I can't see all changes by navigating in a per commit basis, I keep looking at  the outdate commits and the only way to see the updated ones is to see everything in one page (File Changes).

2. The commit msgs could be more helpful, since it's more going on than it's descrived :)

aodntools/timeseries_products/common.py
```diff
@@ -4,3 +4,128 @@
...
+
+    :param nc: open xarray dataset
+    :param site_code: code of the mooring site
+    :param variables_of_interest: variable name or list of names to be included in the product
+                                  (the string 'velocity' implies checks specific to velocity files,
+                                  with variabls 'UCUR' AND 'VCUR' required)
+    :return: list of failed tests
+    """
+
+    if variables_of_interest == 'velocity':
```
ocehugo, line 77: I would avoid this. You are mangling too many definitions an already multi-type argument (list or string).

IMO, it's better to be explicit on definitions before the call:

```
velocity = {'UCUR',...}
required_vars = {...}
allowed_dims = {...}
check_file(nc,site_code,velocity,required_vars=required_vars,allowed_dims=allowed_dims)
```

aodntools/timeseries_products/velocity_aggregated_timeseries.py
ocehugo, line 78: See previous comment about the velocity argument. Note the varlist is defined in this function but not used anymore.

aodntools/timeseries_products/common.py
```diff
@@ -128,4 +124,34 @@ def check_file(nc, site_code, variables_of_interest):
     # check qc flag conventions for VoI and depth/pressure
     error_list.extend(check_imos_flag_conventions(nc))
 
-    return error_list
\ No newline at end of file
+    return error_list
+
+
+def check_velocity_file(nc, site_code,
+                        required_variables=('TIME', 'DEPTH', 'LATITUDE', 'LONGITUDE', 'UCUR', 'VCUR'),
```
ocehugo, line 129: WCUR missing?

ocehugo: I don't know if it is because I'm using lambdas heavily lately, but this could be a lambda if it's just a wrapper. This would avoid  that monster of docstring (since most of it is already/should be within check_file fun).

aodntools/timeseries_products/common.py
ocehugo: yeap, or that.

aodntools/timeseries_products/common.py
ocehugo: Yes, but your msg will be out of order because of the set. In the error event, I would expect to see errors in a per variable fashion and in order of importance (variable not in file > missing_qc > general conventions). Using a list and moving the accepted_conventions out of the loop solve the out of order and  over-reporting problem!

ocehugo:
@mhidas , will leave to you if you want to follow up on my comments or just merge

aodntools/timeseries_products/common.py
mhidas: > WCUR missing?

No, WCUR is not required. Although it _is_ actually there in just about every file, we explicitly said in the plan and the [documentation](https://github.com/aodn/python-aodntools/blob/master/aodntools/timeseries_products/Documentation/velocity_aggregated_timeseries.md#input) that only UCUR and VCUR are absolutely required. I'm no longer sure why we did this, but that was the decision. Eduardo added [code](https://github.com/aodn/python-aodntools/blob/b68c421ce1258460dabaa7b6cc9c48cdfc0de1bb/aodntools/timeseries_products/velocity_aggregated_timeseries.py#L205) to handle the case where WCUR is missing.

mhidas:
Thanks @ocehugo I appreciate your comments, but I don't want to spend any more time on this PR, so I'll merge it.

PR #141, opened 2021-06-10
GitHub Actions to test against python 3.8

mhidas:
Update the GitHub Actions config to test against the most relevant versions of python, in line with python-aodncore.


lbesnard:
@mhidas do you wnat to keep this as a draft?

mhidas:
It's good to go!

Issue #46, opened 2019-06-17
float type 16 not recognised


lbesnard:
The template suggests that it is possible to use 'float16'
https://github.com/aodn/aodn-netcdf-tools/blob/c62ba047f9aec92c340c71b444b9fe31250feb37/ncwriter/template_schema.json#L57

However, the netcdf-writer bugs for me for float16. (float, float32 and float64 work fine)
```
  File "/home/github_repo/chef/src/python-aodndata/python-aodndata-virtualenv/local/lib/python2.7/site-packages/ncwriter/template.py", line 305, in create_variables
    ncvar = self.ncobj.createVariable(varname, **var_c_opts)
  File "netCDF4/_netCDF4.pyx", line 2557, in netCDF4._netCDF4.Dataset.createVariable
  File "netCDF4/_netCDF4.pyx", line 3490, in netCDF4._netCDF4.Variable.__init__
TypeError: illegal primitive data type, must be one of ['i8', 'f4', 'f8', 'S1', 'i2', 'i4', 'u8', 'u4', 'u1', 'u2', 'i1'], got float16


```

ocehugo:
AFAICR, there is no exact representation for float16 in C. The closest equivalent is ```short``` but a bit better than float16.

ggalibert:
I'm not too fussed about supporting float 16, we barely ever use them. Float 32 though is more important.

mhidas:
Ok, my bad. It turns out this is "false advertising" in the template schema. NetCDF actually doesn't support a 16-bit floating point type (see https://www.unidata.ucar.edu/software/netcdf/docs/data_type.html).

So the fix is to review all the values specified [there](https://github.com/aodn/aodn-netcdf-tools/blob/c62ba047f9aec92c340c71b444b9fe31250feb37/ncwriter/template_schema.json#L55), and remove any that are actually not consistent with netCDF.

PR #143, opened 2021-08-03
(Fix) https://github.com/aodn/python-aodntools/issues/46

lbesnard:
@mhidas, is that all that needs to be done?

### DIFF
--- a/aodntools/ncwriter/template_schema.json
+++ b/aodntools/ncwriter/template_schema.json
@@ -54,7 +54,7 @@
                     "type": "string",
                     "enum": [
                         "int", "int8", "int16", "int32", "int64",
-                        "float", "float16", "float32", "float64", "double",
+                        "float", "float32", "float64", "double",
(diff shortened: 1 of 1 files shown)


mhidas:
> @mhidas, is that all that needs to be done?

That fixes it for data types specified as "float16". However, as I [suggested](https://github.com/aodn/python-aodntools/issues/46#issuecomment-503903310), a complete fix would include reviewing the full list in that enum and removing all type specifications that are not supported in netCDF. I think the other "intN" and "floatN" type should be fine, but not sure about some of the single-character codes.

Probably the easiest way to do this is to try creating a NetCDF variable with each of those types and see if you get the error

lbesnard:
> the full list in that enum and removing all type specifications that

I did that and everything else seemed to be valid

mhidas:
:+1: Great!

mhidas (approved):

Issue #13, opened 2018-10-11
Allow attribute type to be specified in JSON


mhidas:
Currently global & variable attributes are just specified as key:value pairs with the value either a string, a  list, or a number, to be converted into netCDF atttributes by the netCDF4 library's `setncattr` and `setncatts` methods. This is fine in most cases, except when we explicitly want to set the type of an attribute to e.g. double (e.g. to match the data type of the variable). 

Within Python code it's easy to specify the attribute type by  setting its value to the appropriate `numpy` object (e.g. `np.float32(1.234)`). However, when numeric values are specified in a JSON template, they are automatically converted into int, long, or  float (https://docs.python.org/2/library/json.html#json-to-py-table).  We need to allow attribute values in the template to be specified as another (JSON) object, with properties "type" and "data".

mhidas:
> My preference is that any numeric attribute that is related to a particular variable should have the same type.

There is one tricky possibility: what if you want an integer attribute for a floating-point variable? (e.g. some kind of overall status flag)

> For global attributes, things like geospatial_lat_min/max should have the same type as the latitude variable for example.

I'm not sure that's necessary, but would automatically be the case if we do #17 .

> example with wmo codes in global attributes. You want them as integer. But they could be falsely written as floats in a NetCDF

This can already be done via the JSON. If a number is specified with no decimal point, it's read as an integer.

mhidas:
Actually, following on from my last comment above, the rule for variable attributes could be to cast all _floating point_ attributes to the variable's type.

By the way, remember that this is only an issue for values specified in a JSON template. Attribute types can be easily controlled in Python code.

ocehugo:
My simpleton argument was too shallow. my point is this feature is unnecessary.

First, we should avoid automatic internal casting/conversion, because the json python package would do the right thing (it convert/recast automatically).

I can see a point in providing a string in the template (something like "np.float32(1234.0005)"), but this would create code boilerplate, eval calls, and will change the precision unnecessarily. As said, python json package already cast to the correct precision.

Maybe this issue is due to a confusion regarding type names:

py.float is actually np.double because the py.float is actually a c double.  Hence, json can read double, float and ints in any precision up to 64. I can hardly see anyone using float128 and remember how many digits they have to type in the template to match the exact precision. We should dump json format anyway in this case...

> You do care actually, one example is the use of CF packed data.

[...]

PR #144, opened 2021-08-03
(Feat) fix https://github.com/aodn/python-aodntools/issues/47 https://github.com/aodn/python-aodntools/issues/13 - autoconvert some var attr

### LINKED ISSUE #47: var attr seem to have different type as variable datatype definition
template:
```json
"CPHL": {
      "_dimensions": ["TIME"],
      "_datatype": "float32",
      "long_name": "mass_concentration_of_inferred_chlorophyll_from_relative_fluorescence_units_in_sea_water_concentration_of_chlorophyll_in_sea_water",
      "comment": "spiritCountToEngPP.m: artificial chlorophyll data derived from WetLabs ECO-FLNTU bio-optical sensor raw counts measurements using calibration_formula. Originally expressed in ug/l, 1l = 0.001m3 was assumed.",
      "units": "mg m-3",
      "_FillValue": 999999.0,
      "valid_min": 0.0,
      "valid_max": 3.0,
      "coordinates": "TIME LATITUDE LONGITUDE NOMINAL_DEPTH"
    }
```

Creating the above variable with a datatype other that float64, creates an issue with the compliance checker

```bash
CPHL:           Attribute 'valid_min' should have same type as variable.
                Attribute 'valid_max' should have same type as variable.
```

lbesnard:
autoconversion of _FillValue, valid_min, valid_max, valid_range as specified in
https://github.com/aodn/python-aodntools/issues/13#issuecomment-432006164

@leonardolaiolo FYI,
under python-aodntools/test_aodntools/ncwriter, to run the new unittests I do the following
```bash
pytest -k 'test_var_attr_'
```

close #47 #13  once merged

### DIFF
--- a/aodntools/ncwriter/template.py
+++ b/aodntools/ncwriter/template.py
@@ -295,6 +295,11 @@ class DatasetTemplate(NetCDFGroupDict):
         **kwargs are included here to overload all options for all variables
         like `zlib` and friends.
         """
+
+        # variable attributes to convert to the same type as the variable
+        # datatype
+        varattrs_to_convert_to_datatype = ['_FillValue', 'valid_min', 'valid_max', 'valid_range']
+
         for varname, varattr in self.variables.items():
             if not varattr['_dimensions']:  # no kwargs in createVariable
                 ncvar = self.ncobj.createVariable(varname, varattr['_datatype'])
@@ -307,6 +312,11 @@ class DatasetTemplate(NetCDFGroupDict):
             # add variable values
             if varattr['_data'] is not None:
                 ncvar[:] = varattr['_data']
+
+            # convert some variables attribute to variable datatype
+            for varattr_to_convert in varattrs_to_convert_to_datatype:
+                if varattr_to_convert in varattr.keys():
+                    varattr[varattr_to_convert] = np.array(varattr[varattr_to_convert]).astype(varattr['_datatype']).item()
 
             # add variable attributes
             ncvar.setncatts(metadata_attributes(varattr))
--- a/test_aodntools/ncwriter/test_template.py
+++ b/test_aodntools/ncwriter/test_template.py
@@ -348,12 +348,16 @@ class TestDataValues(TemplateTestCase):
                 'X': {
                     '_dimensions': ['TIME'],
                     '_datatype': 'float32',
+                    'valid_min': 1,
+                    'valid_max': 5,
                     '_FillValue': -999.,
                     '_data': self.data_array
                 },
                 'Y': {
                     '_dimensions': ['TIME'],
                     '_datatype': 'float32',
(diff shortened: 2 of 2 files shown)


test_aodntools/ncwriter/test_template.py
```diff
@@ -348,12 +348,16 @@ def setUp(self):
...
+                    'valid_min': 1,
+                    'valid_max': 5,
                     '_FillValue': -999.,
                     '_data': self.data_array
                 },
                 'Y': {
                     '_dimensions': ['TIME'],
                     '_datatype': 'float32',
+                    'valid_min': -4,
+                    'valid_max': 5,
```
mhidas, line 360: You're testing the exact same case 4 times (variable is `float32`, attribute looks like integer). How about changing one of the variables to `float64`, and maybe adding an `int32`, and specifying some of these numbers with a decimal point.

Should also test the correct handling of 
* 'valid_range', as it is array-valued (e.g. `'valid_range': [-4, 5]`), and
* '_FillValue', which is a special attribute, handled by the NetCDF4 library.

aodntools/ncwriter/template.py
```diff
@@ -295,6 +295,11 @@ def create_variables(self, **kwargs):
         **kwargs are included here to overload all options for all variables
         like `zlib` and friends.
         """
+
+        # variable attributes to convert to the same type as the variable
+        # datatype
+        varattrs_to_convert_to_datatype = ['_FillValue', 'valid_min', 'valid_max', 'valid_range']
```
mhidas, line 301: Don't need to include '_FillValue' as it is handled by the NetCDF4 library when it creates the variable.

aodntools/ncwriter/template.py
```diff
@@ -308,6 +313,11 @@ def create_variables(self, **kwargs):
             if varattr['_data'] is not None:
                 ncvar[:] = varattr['_data']
 
+            # convert some variables attribute to variable datatype
+            for varattr_to_convert in varattrs_to_convert_to_datatype:
+                if varattr_to_convert in varattr.keys():
+                    varattr[varattr_to_convert] = np.array(varattr[varattr_to_convert]).astype(varattr['_datatype']).item()
```
mhidas, line 319: This looks more complicated than it needs to be, and doesn't work for array-valued attributes. This seems to work for me:
```suggestion
                    varattr[varattr_to_convert] = np.array(varattr[varattr_to_convert], dtype=varattr['_datatype'])
```

mhidas (changes requested):
A couple of suggestions to make the tests more complete, and to make sure they pass.

lbesnard:
are we all good to merge this ?

mhidas (approved):
:+1:

Issue #112, opened 2020-03-18
Fix deprecation and run-time warnings


mhidas:
Various `DeprecationWarning` and `RuntimeWarning` pop up when running the unittests. These should ideally be fixed before they become errors.

E.g. 
```
test_aodntools/ncwriter/test_template.py::TestDatasetTemplate::test_ensure_completeness
  /home/travis/build/aodn/python-aodntools/test_aodntools/ncwriter/test_template.py:282: DeprecationWarning: Please use assertRaisesRegex instead.
    template.ensure_completeness)
  /home/travis/build/aodn/python-aodntools/test_aodntools/ncwriter/test_template.py:286: DeprecationWarning: Please use assertRaisesRegex instead.
    template.ensure_completeness)

test_aodntools/ncwriter/test_template.py::TestDatasetTemplate::test_ensure_consistency
  /home/travis/build/aodn/python-aodntools/test_aodntools/ncwriter/test_template.py:318: DeprecationWarning: Please use assertRaisesRegex instead.
    template.ensure_consistency)  # now should fail because dim X is already set
  /home/travis/build/aodn/python-aodntools/test_aodntools/ncwriter/test_template.py:323: DeprecationWarning: Please use assertRaisesRegex instead.
    self.assertRaisesRegexp(ValidationError, 'undefined dimensions', template.ensure_consistency)
  /home/travis/build/aodn/python-aodntools/test_aodntools/ncwriter/test_template.py:330: DeprecationWarning: Please use assertRaisesRegex instead.
    template.ensure_consistency

test_aodntools/ncwriter/test_template.py::TestDatasetTemplate::test_invalid_json
  /home/travis/build/aodn/python-aodntools/test_aodntools/ncwriter/test_template.py:105: DeprecationWarning: Please use assertRaisesRegex instead.
    self.assertRaisesRegexp(ValueError, error_pattern, DatasetTemplate.from_json, BAD_JSON)

test_aodntools/timeseries_products/test_aggregated_timeseries.py::TestAggregatedTimeseries::test_main_aggregator
  /home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/numpy/core/fromnumeric.py:83: RuntimeWarning: invalid value encountered in less
    return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
  /home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/numpy/core/fromnumeric.py:83: RuntimeWarning: invalid value encountered in greater
[...]

PR #148, opened 2022-01-26
Use assertRaisesRegex instead of assertRaisesRegexp for Python 3.11 compatibility.

tirkarthi:
Related to https://github.com/aodn/python-aodntools/issues/112 . The deprecation warnings are now error in Python 3.11.


alexgleith:
Can we merge this @mhidas?

PR #154, opened 2022-05-04
set GH Actions to test against python 3.5.2

mhidas:


mhidas:
This is of no use - Python 3.5.2 is not available. The closest is 3.5.4, and under that version the unittests pass, while one still fails under 3.5.2 :face_with_spiral_eyes:

PR #156, opened 2022-05-11
fix: mooring product processing level attribute for level 2

evacougnon:
@mhidas the `file_version` and `file_version_quality_control` description in the hourly timeseries products did not match our [documentation](https://s3-ap-southeast-2.amazonaws.com/content.aodn.org.au/Documents/IMOS/Conventions/IMOS_NetCDF_Conventions.pdf)

The aggregated product showing `Level 1` file version has the same description as shown in the table on page 25, so I assumed `Level 2` should also show the corresponding text. 

Following this fix we may want to re-process the hourly timeseries


mhidas (approved):
:+1: Good catch, and thanks for the fix.
While we're at it, though, the same fix is also needed in the gridded product.

mhidas (approved):

PR #165, opened 2022-06-15
Fix velocity hourly IndexError

mhidas:
This is an error that occurs when the `time_deployment_start/end` attributes in an input file represent a longer time span than what is actually covered by the data. Previously this could lead to the code trying to select and process an empty subset of the data.


mhidas:
@bpasquer @lbesnard My usual reviewer Leo is away, but this is just a very quick fix that should be pretty easy to review based on my explanation above. Any chance one of you could do a quick review?

bpasquer (approved):

mhidas:
Thanks @bpasquer !

PR #169, opened 2022-06-29
Fix a failing test, which was failing due to modifying a list in plac…

alexgleith:
Added a little test case to test single files. Discovered the issue is in the `main_aggregator` function modifying a list in place. Could be unique to my Python version, but wanted to send it in anyway.

Most of these changes are formatting... sorry about that.

### DIFF
--- a/aodntools/timeseries_products/aggregated_timeseries.py
+++ b/aodntools/timeseries_products/aggregated_timeseries.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import argparse
+from copy import deepcopy
 import json
 import os
 import shutil
@@ -199,6 +200,7 @@ def main_aggregator(files_to_agg, var_to_agg, site_code, input_dir='', output_di
     :param opendap_url_prefix: URL prefix for OPENAP access (to be prepended to paths in files_to_agg)
     :return: name of the resulting file, list of rejected files
     """
+    files_to_agg = deepcopy(files_to_agg)
 
     time_units="days since 1950-01-01 00:00:00 UTC"
     time_calendar="gregorian"
--- a/test_aodntools/timeseries_products/test_aggregated_timeseries.py
+++ b/test_aodntools/timeseries_products/test_aggregated_timeseries.py
@@ -43,44 +43,82 @@ class TestAggregatedTimeseries(BaseTestCase):
             assert len(result) == 0
 
     def test_main_aggregator(self):
-        output_file, bad_files = main_aggregator(INPUT_FILES, 'TEMP', 'NRSROT', input_dir=TEST_ROOT,
-                                                 output_dir='/tmp')
+        output_file, bad_files = main_aggregator(
+            TEST_FILES,
+            VAR_TO_AGG,
+            SITE_CODE,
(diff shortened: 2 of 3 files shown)


alexgleith:
I've split this into two commits @mhidas

aodntools/timeseries_products/aggregated_timeseries.py
```diff
@@ -199,6 +200,7 @@ def main_aggregator(files_to_agg, var_to_agg, site_code, input_dir='', output_di
     :param opendap_url_prefix: URL prefix for OPENAP access (to be prepended to paths in files_to_agg)
     :return: name of the resulting file, list of rejected files
     """
+    files_to_agg = deepcopy(files_to_agg)
```
mhidas, line 203: :+1: Ok, this look like the right thing to do, since we're then using `files_to_agg.remove()` below to cull the list. However, 
* this doesn't seem to cause any test to fail for me (in python 3.8.13); and
* the same thing is done in [velocity_aggregated_timeseries.py](https://github.com/aodn/python-aodntools/blob/9098f6a0092289c56da2c4eba200af1e55e048ff/aodntools/timeseries_products/velocity_aggregated_timeseries.py#L87) and that doesn't fail for you?

test_aodntools/timeseries_products/test_aggregated_timeseries.py
```diff
@@ -7,24 +7,41 @@
...
 
 
 class TestAggregatedTimeseries(BaseTestCase):
     EXPECTED_OUTPUT_FILE = os.path.join(
-        TEST_ROOT, 'IMOS_ANMN-NRS_TZ_20181213_NRSROT_FV01_TEMP-aggregated-timeseries_END-20190523_C-20220607.nc'
+        TEST_ROOT,
+        "IMOS_ANMN-NRS_TZ_20181213_NRSROT_FV01_TEMP-aggregated-timeseries_END-20190523_C-20220607.nc",
     )
 
+    def test_one_file(self):
```
mhidas, line 34: I don't think this test has any relevance to the code change you made in  `main_aggregator` as it's not calling that function. This is just the `check_file` function, which is already done [elsewhere](https://github.com/aodn/python-aodntools/blob/9098f6a0092289c56da2c4eba200af1e55e048ff/test_aodntools/timeseries_products/test_common.py#L49).

mhidas (commented):
Good pick up on the modification of function argument. However I don't think the fix is quite complete.

aodntools/timeseries_products/aggregated_timeseries.py
alexgleith: I haven't looked at it carefully, but I think that if you run on the list once, it'll be fine. Running it on it a second time, it might behave weird. This happens in the tests, but probably not in real life...

test_aodntools/timeseries_products/test_aggregated_timeseries.py
alexgleith: I added this test while I was working out what was failing and why.

I'm happy to remove it if you don't think it's useful, but it is a valid test, I think?

mhidas:
What about something as simple as this? #172

PR #172, opened 2022-08-18
Fix functions that modify files_to_agg argument in place

mhidas:
... and add a simple test assertion to confirm fix


alexgleith (approved):
That's a good idea!

Being explicit is nice :-)

PR #176, opened 2023-01-05
Fixing the temp file permission bug, plus Velocity gridded product code

mphemming:
**Fixing the temp file permission bug**

So the issue is that the temp file is still open in Spyder, meaning that shutils cannot move the file at the end of the main aggregator function.

I edited line 214 to include 'fd' as output:
_, temp_outfile = tempfile.mkstemp(suffix='.nc', dir=output_dir) to fd, temp_outfile = tempfile.mkstemp(suffix='.nc', dir=output_dir).

I can then close the object adding the following code to line 355 prior to using 'shutil.move':

os.close(fd)

This now works without an error for both 'aggregated_timeseries.py' and 'velocity_aggregated_timeseries.py'.

**Velocity gridded timeseries product code**

An attempt to create a velocity gridded product using the groupby method, as used to produce the other gridded products. Gridded VCUR, UCUR and WCUR are provided, along with metadata.

The code could probably do with improving - I tried to run the loop in parallel with no luck. Although for one year's worth of velocity data it takes about 4 mins on my laptop without remote access to a server so not too bad.

Attributes are mostly correct, but a few attributes might need tweaking or new ones added. Also keep in mind that the method of adding attributes might not be satisfactory for the AODN. For example, I have copied the abstract, title and lineage attributes into the script rather than relying on the json template.

This product is probably good enough for NSW-IMOS purposes, but we may develop different versions, or improve the code in future. If so, I can create a new pull request.


mphemming:
@mhidas Please do not accept the first commit, I had to delete the nc files in the forked `python-aodntools/tree/master/test_aodntools` in order to clone my fork to my local directory as it complained.

aodntools/timeseries_products/aggregated_timeseries.py
```diff
@@ -349,6 +349,7 @@ def main_aggregator(files_to_agg, var_to_agg, site_code, input_dir='', output_di
                             (var_to_agg + "-" + product_type),
                             ('END-'+ time_end_filename), 'C-' + current_utc_timestamp(DATESTAMP_FORMAT)]) + '.nc'
     ncout_path = os.path.join(output_dir, output_name)
+    os.close(fd)
```
mhidas, line 352: :+1: That makes sense.
In fact, since we don't use this file descriptor to write to the file (instead we just use the `temp_outfile` as a name for a new netCDF4 file - [line 239](https://github.com/aodn/python-aodntools/blob/c3bad45d406bcc19af4ba67431ad8b378c8db114/aodntools/timeseries_products/aggregated_timeseries.py#L239)), you could even close it immediately after the call to `tempfile.mkstemp()`

mhidas (changes requested):
Thanks @mphemming . It's nice to clean up the hadnling of these temp files. 
Note that velocity_hourly_timeseries.py also creates and moves a temp file in a similar way, so perhaps you could apply the same fix to it?

A couple of more general requests for this Pull Request:
* I can only approve/merge or reject the whole PR as it is, not individual commits, so if you can find a way to exclude the deletion of all those test files (your first commit), that would be better. 
* The changes you propose in your third commit are totally unrelated to the second one, so should be a separate Pull Request.

mhidas:
Let me know if you are able to make those changes or if you need help. 
If it's easier I can just pull out your changes from the second commit into a new PR and you can review them?

mphemming:
> Let me know if you are able to make those changes or if you need help. If it's easier I can just pull out your changes from the second commit into a new PR and you can review them?

@mhidas I think I've fixed it? reverted the velocity gridded product code and deletion of NetCDF files. I've also changed the line where I include the new code for the aggregated products (just after tempfile.mkstemp() is used), and now also do the same for the velocity hourly product. Let me know if I need to do anything else. 

I can create a new pull request for the velocity gridded product after this request if you are interested.

mhidas:
Thanks @mphemming . That's better, but still not quite there. If you have a look at the list of [changed files](https://github.com/aodn/python-aodntools/pull/176/files), you can see your velocity gridded code is still in there, and there are still 3 test files missing. It gets a bit messy when you start revering bits & pieces...

Might be easiest to start a new branch and just apply the changes you want. If you're feeling adventurous you could just [cherry-pick](https://docs.github.com/en/desktop/contributing-and-collaborating-using-github-desktop/managing-commits/cherry-picking-a-commit) the 3 commits here that you need. In fact I just did that and created a draft PR for you #178 -- if you like we can just review and merge that one.

mhidas:
Closing this in favour of #178, with the velocity gridded code to come in a future PR

PR #182, opened 2025-09-09
7287 flag values convert to bytes

mphemming:
I also tested creating a LTSP and worked as expected. 

The failing pytest was due to `xarray==2025.9.1`. When testing locally in my virtual environment it passed with `xarray==2025.3.1`. I changed the setup.py to reflect this, but GH actions don't seem to run when the base is the `V2` branch.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,9 +2,9 @@ name: python-aodntools
 
 on:
   push:
-    branches: [ master ]
+    branches: [ master, v2 ]
   pull_request:
-    branches: [ master ]
+    branches: [ master, v2 ]
 
 jobs:
   build:
--- a/aodntools/ncwriter/template.py
+++ b/aodntools/ncwriter/template.py
@@ -298,7 +298,7 @@ class DatasetTemplate(NetCDFGroupDict):
 
         # variable attributes to convert to the same type as the variable
         # datatype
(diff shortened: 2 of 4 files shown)


mhidas:
> but GH actions don't seem to run when the base is the `V2` branch.

It's configured to run only on PR and push into the `master` branch (see lines 5 & 7 of the [workflow](https://github.com/aodn/python-aodntools/blob/36938d0846799e80f10b2dcb5d73040b328b4cb5/.github/workflows/test.yml#L5))

Since we're now effectively making `v2` the new main branch, we should add:
```yaml
on:
  push:
    branches: [ master, v2 ]
  pull_request:
    branches: [ master, v2 ]
```

mphemming:
> > but GH actions don't seem to run when the base is the `V2` branch.
> 
> It's configured to run only on PR and push into the `master` branch (see lines 5 & 7 of the [workflow](https://github.com/aodn/python-aodntools/blob/36938d0846799e80f10b2dcb5d73040b328b4cb5/.github/workflows/test.yml#L5))
> 
> Since we're now effectively making `v2` the new main branch, we should add:
> 
> ```yaml
> on:
>   push:
>     branches: [ master, v2 ]
>   pull_request:
>     branches: [ master, v2 ]
> ```

that worked, thanks!

mhidas (approved):

